PR #6, opened 2015-11-19
Add ComputedStylePropertyMap and StylePropertyMaps

wilddamon:


src/computed-style-property-map.js
```diff
@@ -0,0 +1,44 @@
...
+
+ComputedStylePropertyMap.prototype.get = function(property) {
+  if (typeof property != 'string') {
+    throw new TypeError('parameter 1 is not of type \'string\'');
+  }
+  // TODO: define types
+  throw new TypeError('Not implemented yet');
+};
+
+function getComputedStyleMap(element) {
```
rjwright: This should be a method on document, right? Also, do we want to make a new one each time or do we want to cache them (maybe on the element?)

PR #8, opened 2015-11-23
Add testing & loading configs

rjwright:
Using Mocha and [Chai asserts](http://chaijs.com/api/assert/). Later I might add support for grunt test using Karma.

You can run the tests by doing

```
sudo apt-get install node
npm install -g mocha
npm install -g chai
```

then opening typed-om/test/runner.html in a web browser.

Users can now use the code by putting

```
<script src='typed-om/typed-om.js'></script>
```

in their HTML code.

This PR just contains a very simple example test for now. Will need to add actual tests for NumberValue.


rjwright:
@wilddamon @shans

src/scope.js
```diff
@@ -15,4 +15,5 @@
 
 typedOM = {};
 baseClasses = {};
-
+if (!TYPED_OM_TESTING)
+	var typedOMTesting = null;
```
shans, line 21: don't you need to set typedOMTesting to an empty object (or to window) if TYPED_OM_TESTING is true?

shans:
LGTM, assuming it works :)

You should put the documentation above into the project's README.md.

rjwright:
Yeah it works. Most of the instructions are already in the readme, but I'll ad the link to Chai.

PR #10, opened 2015-11-24
Add some more test cases for NumberValue

rjwright:


rjwright:
@shans @wilddamon ptal

PR #12, opened 2015-11-24
Add computed style map testing + tweaks

rjwright:
Take ComputedStylePropertyMap off window as it is not part of the API. Expose it to testing instead.

Expose StylePropertyMap to testing.

Add test cases for ComputedStylePropertyMap.


src/style-property-map.js
```diff
@@ -13,20 +13,22 @@
...
-}
+	StylePropertyMap.prototype = {
+	  append: function(property, value) {},
+	  delete: function(property) {},
+	  get: function(property) {},
+	  getAll: function(property) {},
+	  has: function(property) {},
+	  set: function(property, value) {},
+	  getProperties: function() {}
+	}
```
rjwright, line 28: The only change to this block is indenting.

rjwright:
@shans @wilddamon

PR #15, opened 2015-11-26
Add CalcLength

nattridgell:
Adds:
- CalcLength constructor
- A couple of tests for CalcLength
- fromValue / fromDictionary in LengthValue calls
  corresponding constructors for SimpleLength / CalcLength
- extremely simple tests for fromValue / fromDictionary
- remove unnecessary tags for consistency
  (@constructor, @enum tags)

### DIFF
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -20,27 +20,26 @@
    */
   function CalcLength(dictionary) {
     if (typeof dictionary != 'object') {
-      throw(new TypeError('CalcLength must be passed a dictionary object'));
+      throw new TypeError('CalcLength must be passed a CalcDictionary object.');
     }
+
+    var hasValidLength = false;
 
     // NOTE: Does not check dictionary for invalid fields.
     for (var type in LengthValue.LengthType) {
       var value = dictionary[type];
       if (typeof value == 'number') {
         this[type] = value;
-      } else if (typeof value == 'string') {
-        var nValue = Number.parseFloat(value);
-        if (!isNaN(nValue)) {
-          this[type] = nValue;
-        } else {
-          throw(new TypeError('Value of each field must be null, a number or a numeric string.'));
-        }
+        hasValidLength = true;
       } else if (value == undefined || value == null) {
         this[type] = null;
       } else {
-        throw(new TypeError('Value of each field must be null, a number or a numeric string.'));
+        throw new TypeError('Value of each length must be null or a number.');
       }
     }
+
+    if (!hasValidLength)
+      throw new TypeError('A CalcDictionary must have at least one valid length.');
   }
 
   CalcLength.prototype = shared.LengthValue.prototype;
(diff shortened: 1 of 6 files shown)


src/calc-length.js
```diff
@@ -0,0 +1,52 @@
...
+      throw(new TypeError('CalcLength must be passed a dictionary object'));
+    }
+
+    // NOTE: Does not check dictionary for invalid fields.
+    for (var type in LengthValue.LengthType) {
+      var value = dictionary[type];
+      if (typeof value == 'number') {
+        this[type] = value;
+      } else if (typeof value == 'string') {
+        var nValue = Number.parseFloat(value);
```
wilddamon: Need to check with Shane if we care about parsing strings here

test/js/length-value.js
```diff
@@ -0,0 +1,13 @@
+suite('LengthValue', function() {
+  test('fromValue returns instanceof a LengthValue and StyleValue', function() {
```
wilddamon: Should these also check for SimpleLength and CalcLength types here?

src/calc-length.js
nattridgell: I'll remove it for now and then we can add it back in later if need be.

nattridgell:
I hope this is what you meant :-)

src/calc-length.js
```diff
@@ -0,0 +1,51 @@
...
+   * TODO: CalcLength(cssString)
+   */
+  function CalcLength(dictionary) {
+    if (typeof dictionary != 'object') {
+      throw new TypeError('CalcLength must be passed a CalcDictionary object.');
+    }
+
+    var hasValidLength = false;
+
+    // NOTE: Does not check dictionary for invalid fields.
```
wilddamon: Please add a comment saying the check is done in fromDictionary.

src/length-value.js
```diff
@@ -65,11 +64,44 @@
...
+        firstValidType = type;
+      } else if (value != null && value != undefined) {
+        throw new TypeError('Value of each length must be null or a number.');
+      }
+    }
+
+    if (numLengthsFound < 1) {
+      throw new TypeError('A CalcDictionary must have at least one valid length.');
+    } else if (numLengthsFound < 2) {
+      return new SimpleLength(dictionary[firstValidType], firstValidType);
```
wilddamon: As discussed, don't do this for consistency with calc length constructor - a calc with a single value is ok.

src/length-value.js
```diff
@@ -65,11 +64,44 @@
...
+    } else {
+      return new CalcLength(dictionary);
```
wilddamon: As discussed, we don't want to do this any more - you can move all of this logic into the constructor, and get rid of the part where you create a SimpleLength.
Sorry about that...

wilddamon:
LGTM

PR #16, opened 2015-11-27
Add cssString attribute to SimpleLength

nattridgell:
- Adds the cssString attribute to SimpleLength.
- Changed string value of percent in enum, such that
  SimpleLength(10, 'percent').cssString is '10%'
- Add some tests

### DIFF
--- a/src/simple-length.js
+++ b/src/simple-length.js
@@ -23,7 +23,7 @@
   function SimpleLength(value, type) {
     if (arguments.length == 2 && typeof type == 'string' &&
       type.toUpperCase() in LengthValue.LengthType) {
-      this.type = LengthValue.LengthType[type.toUpperCase()];
+      this.type = type.toUpperCase();
       if (typeof value == 'number') {
         this.value = value;
       } else if (typeof value == 'string') {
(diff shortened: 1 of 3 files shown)


test/js/simple-length.js
```diff
@@ -17,12 +17,26 @@ suite('SimpleLength', function() {
   test('SimpleLength constructor works correctly for numbers and numeric strings', function() {
     var valueFromString;
     assert.doesNotThrow(function() {valueFromString = new SimpleLength('9.2', 'px')});
-    assert.strictEqual(valueFromString.type, 'px');
+    assert.strictEqual(valueFromString.type, 'PX');
```
wilddamon: Is this case change intentional?

test/js/simple-length.js
```diff
@@ -17,12 +17,26 @@ suite('SimpleLength', function() {
...
     assert.strictEqual(valueFromString.value, 9.2);
 
     var valueFromNumber;
     assert.doesNotThrow(function() {valueFromNumber = new SimpleLength(10, 'px')});
-    assert.strictEqual(valueFromNumber.type, 'px');
+    assert.strictEqual(valueFromNumber.type, 'PX');
```
wilddamon: Same here...

src/simple-length.js
```diff
@@ -23,7 +23,7 @@
   function SimpleLength(value, type) {
     if (arguments.length == 2 && typeof type == 'string' &&
       type.toUpperCase() in LengthValue.LengthType) {
-      this.type = LengthValue.LengthType[type.toUpperCase()];
+      this.type = type.toUpperCase();
```
wilddamon, line 26: You can't really assume that the string for type can be uppercased into an enum value, especially now that you're changing the percent string to "%".

You may need to define some helper functions that do the conversion back and forth outside of this class. This will need to involve some switch statements and/or a map. In C++ we do this: https://codereview.chromium.org/1408033008/diff/20001/third_party/WebKit/Source/core/css/LengthValue.cpp

PR #19, opened 2015-11-29
Add CalcLength.cssString

nattridgell:
- Creates the cssString attribute for calcLength
  in the constructor.
- Create a function in LengthValue to return the cssString
  representation of each LengthValue.LengthType
  [Note: uses switch/case block for extensibility](LengthValue.cssStringTypeRepresentation)
- Add appropriate cssString test cases
  (single, negative, multi, percent)

### DIFF
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -24,6 +24,19 @@
     'cm', 'mm', 'q', 'in', 'pc', 'pt'
   ];
 
+  LengthValue.cssStringTypeRepresentation = function(type) {
+    if (LengthValue.LengthType.indexOf(type) < 0) {
+      throw new TypeError('Invalid LengthType.');
+    }
+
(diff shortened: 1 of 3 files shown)


src/length-value.js
```diff
@@ -24,6 +24,19 @@
     'cm', 'mm', 'q', 'in', 'pc', 'pt'
   ];
 
+  LengthValue.cssStringTypeRepresentation = function(type) {
+    if (LengthValue.LengthType.indexOf(type) < 0) {
+      throw new TypeError('Invalid LengthType.');
+    }
+
+    switch (type) {
```
nattridgell, line 32: This switch statement is a bit excessive, but I thought it would fit with the code style in ComputedStylePropertyMap, plus easy / clear how/where to add more cases.

PR #21, opened 2015-12-01
CalcLength multiply method

njlawton:
Implemented method to multiply CalcLength objects by a multiple

### DIFF
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -66,7 +66,17 @@
 
   // Length Calculation Methods
   CalcLength.prototype.multiply = function(multiplier) {
-    return new CalcLength(null);
+    var calcDictionary = new Object;
+
+    //iterate through all length types and multiply all non null lengths 
+    for(var index in shared.LengthValue.LengthType){
+      var type = shared.LengthValue.LengthType[index];
+      if(this[type] != null){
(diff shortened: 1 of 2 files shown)


src/calc-length.js
```diff
@@ -64,6 +64,21 @@
 
   CalcLength.prototype = Object.create(shared.LengthValue.prototype);
 
+  // Length Calculation Methods
+  CalcLength.prototype.multiply = function(multiplier) {
+    var calcDictionary = new Object;
+
+    //iterate through all length types and multiply all non null lengths 
+    for(var index in shared.LengthValue.LengthType){
```
wilddamon: It's probably less confusing (and more common) to write

for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
var type = shared.LengthValue.LengthType[i];
...etc...
}

src/calc-length.js
```diff
@@ -64,6 +64,21 @@
 
   CalcLength.prototype = Object.create(shared.LengthValue.prototype);
 
+  // Length Calculation Methods
+  CalcLength.prototype.multiply = function(multiplier) {
+    var calcDictionary = new Object;
```
wilddamon: Can just write

var calcDictionary = {};

PR #25, opened 2015-12-01
Add keyword and length properties to map

nattridgell:
- adds a list of properties that take length values
  to ComputedStylePropertyMap
- adds a check for keywords
  NOTE: this may have to change / move around
- adds some basic tests for length and keyword values
  for the map
  NOTE: These tests are currently being skipped as parsing is
  unimplemented.
- adds `auto` to KeywordValue
  NOTE: this implementation is subject to change

### DIFF
--- a/src/computed-style-property-map.js
+++ b/src/computed-style-property-map.js
@@ -37,9 +37,9 @@
     if (!value) {
       return null;
     }
-    if (value == 'inherit') {
-      // TODO: Other keywords
-      throw new TypeError('Not implemented yet');
+
+    if (scope.KeywordValue.isKeywordValue(value)) {
+      return new scope.KeywordValue(value);
     }
 
     // TODO: The rest of the properties once the rest of the StyleValues are defined.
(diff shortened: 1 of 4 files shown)


src/computed-style-property-map.js
```diff
@@ -54,6 +54,29 @@
...
+      case 'min-height':
+      case 'min-width':
+      case 'right':
+      case 'text-indent':
+      case 'top':
+      case 'width':
+        // TODO: support it being any length value.
+        return new scope.SimpleLength(value);
+
+      // These properties only take a length value, but do not take percentage.
```
wilddamon, line 73: // TODO: Validation and implementation for this?

nattridgell, line 73: Yep, will add a TODO :-)  Noted that I should add TODO comments for unimplemented things.

nattridgell:
Need to review for new discussed implementation of keyword values.  (for example, review what happens with auto)

PR #26, opened 2015-12-03
Alter LengthValue `from` constructors

nattridgell:
- LengthValue.fromValue and LengthValue.fromDictionary
  are both static methods.
- Alter tests as appropriate.


src/length-value.js
```diff
@@ -62,15 +62,16 @@
     throw new TypeError('Not implemented yet');
   };
 
-  LengthValue.prototype.parse = function(cssString) {
+  // LengthValue static methods
```
wilddamon, line 65: Static methods should be before methods on the prototype

nattridgell:
Partially already addressed in Nick's commit.

PR #27, opened 2015-12-03
Implemented add and subtract methods for CalcLength class 

njlawton:
Implemented add and subtract as well as additional methods used to implement these methods: _convertToCalcLength and equals method

### DIFF
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -92,12 +92,12 @@
 
   CalcLength.prototype._addCalcLengths = function(addedLength) {
     if (!(addedLength instanceof CalcLength)) {
-      throw new TypeError('Objects not of type Calclength');
+      throw new TypeError('Objects not of type CalcLength');
     } 
     
     var calcDictionary = {};
 
-    //Iterate through all posible length types and add there values
+    //Iterate through all possible length types and add there values
     for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
       if (this[type] == null){
@@ -114,12 +114,12 @@
 
   CalcLength.prototype._subtractCalcLengths = function(subtractedLength) {
     if (!(subtractedLength instanceof CalcLength)) {
-      throw new TypeError('Objects not of type Calclength');
+      throw new TypeError('Objects not of type CalcLength');
     } 
     
     var calcDictionary = {};
 
-    //Iterate through all posible length types and add there values
+    //Iterate through all possible length types and add there values
     for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
       if (this[type] == null && subtractedLength[type] == null) {
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -94,6 +94,22 @@
     return this;
   };
 
+  CalcLength.prototype.equals = function(lengthValue) {
+    if (!(lengthValue instanceof CalcLength)){
+      return false;
+    }
+
+    //iterate through all length types and check that both objects contain the same values
+    for(var index in shared.LengthValue.LengthType){
+      var type = shared.LengthValue.LengthType[index];
+      if(this[type] != lengthValue[type]){
+        return false;
+      }
+    }
+    
+    return true;
+  };
+
   scope.CalcLength = CalcLength;
   if (TYPED_OM_TESTING)
     testing.CalcLength = CalcLength;
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -65,10 +65,10 @@
   CalcLength.prototype.multiply = function(multiplier) {
     var calcDictionary = {};
 
-    // Iterate through all length types and multiply all non null lengths
-    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    // Iterate through all length types and multiply all non null lengths 
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
-      if(this[type] != null){
+      if (this[type] != null){
         calcDictionary[type] = this[type] * multiplier;
       }
     }
@@ -79,10 +79,10 @@
   CalcLength.prototype.divide = function(divider) {
     var calcDictionary = {};
 
-    // Iterate through all length types and divide all non null lengths
-    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    // Iterate through all length types and divide all non null lengths 
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
-      if(this[type] != null){
+      if (this[type] != null){
         calcDictionary[type] = this[type] / divider;
       }
     }
@@ -97,8 +97,8 @@
     
     var calcDictionary = {};
 
-    //Iterate through all possible length types and add there values
-    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    // Iterate through all possible length types and add their values
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
       if (this[type] == null){
         calcDictionary[type] = addedLength[type];
@@ -119,8 +119,8 @@
     
     var calcDictionary = {};
 
-    //Iterate through all possible length types and add there values
-    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    // Iterate through all possible length types and add their values
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
       if (this[type] == null && subtractedLength[type] == null) {
         calcDictionary[type] = null;
@@ -145,10 +145,10 @@
       return false;
     }
 
-    //iterate through all length types and check that both objects contain the same values
-    for(var index in shared.LengthValue.LengthType){
+    // Iterate through all length types and check that both objects contain the same values
+    for (var index in shared.LengthValue.LengthType){
       var type = shared.LengthValue.LengthType[index];
-      if(this[type] != lengthValue[type]){
+      if (this[type] != lengthValue[type]){
         return false;
       }
     }
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -92,7 +92,7 @@
 
   CalcLength.prototype._addCalcLengths = function(addedLength) {
     if (!(addedLength instanceof CalcLength)) {
-      throw new TypeError('Objects not of type CalcLength');
+      throw new TypeError('Argument must be a CalcLength');
     }
 
     var calcDictionary = {};
@@ -114,7 +114,7 @@
 
   CalcLength.prototype._subtractCalcLengths = function(subtractedLength) {
     if (!(subtractedLength instanceof CalcLength)) {
-      throw new TypeError('Objects not of type CalcLength');
+      throw new TypeError('Argument must be a CalcLength');
     }
 
     var calcDictionary = {};
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -90,6 +90,10 @@
     return new CalcLength(calcDictionary);
   };
 
+  LengthValue.prototype._convertToCalcLength = function() {
+    return this;
+  };
+
   scope.CalcLength = CalcLength;
   if (TYPED_OM_TESTING)
     testing.CalcLength = CalcLength;
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -138,7 +138,7 @@
     return new CalcLength(calcDictionary);
   };
 
-  LengthValue.prototype._asCalcLength = function() {
+  CalcLength.prototype._asCalcLength = function() {
     return this;
   };
 
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -100,7 +100,9 @@
     // Iterate through all possible length types and add their values
     for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
       var type = shared.LengthValue.LengthType[i];
-      if (this[type] == null) {
+      if (this[type] == null && addedLength[type] == null) {
+        calcDictionary[type] = null;
+      } else if (this[type] == null) {
         calcDictionary[type] = addedLength[type];
       } else if (addedLength[type] == null) {
         calcDictionary[type] = this[type];
@@ -122,7 +124,9 @@
     // Iterate through all possible length types and add their values
     for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
       var type = shared.LengthValue.LengthType[i];
-      if (subtractedLength[type] == null) {
+      if (this[type] == null && subtractedLength[type] == null) {
+        calcDictionary[type] = null;
+      } else if (subtractedLength[type] == null) {
         calcDictionary[type] = this[type];
       } else if (this[type] == null) {
         calcDictionary[type] = -subtractedLength[type];
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -148,8 +148,8 @@
     }
 
     // Iterate through all length types and check that both objects contain the same values
-    for (var index in shared.LengthValue.LengthType) {
-      var type = shared.LengthValue.LengthType[index];
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
+      var type = shared.LengthValue.LengthType[i];
       if (this[type] != other[type]) {
         return false;
       }
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -122,12 +122,10 @@
     // Iterate through all possible length types and add their values
     for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
-      if (this[type] == null && subtractedLength[type] == null) {
-        calcDictionary[type] = null;
+      if (subtractedLength[type] == null) {
+        calcDictionary[type] = this[type];
       } else if (this[type] == null){
         calcDictionary[type] = -subtractedLength[type];
-      } else if (subtractedLength[type] == null) {
-        calcDictionary[type] = -this[type];
       } else {
         calcDictionary[type] = this[type] - subtractedLength[type];
       }
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -65,10 +65,10 @@
   CalcLength.prototype.multiply = function(multiplier) {
     var calcDictionary = {};
 
-    // Iterate through all length types and multiply all non null lengths 
-    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    // Iterate through all length types and multiply all non null lengths
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
       var type = shared.LengthValue.LengthType[i];
-      if (this[type] != null){
+      if (this[type] != null) {
         calcDictionary[type] = this[type] * multiplier;
       }
     }
@@ -79,10 +79,10 @@
   CalcLength.prototype.divide = function(divider) {
     var calcDictionary = {};
 
-    // Iterate through all length types and divide all non null lengths 
-    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    // Iterate through all length types and divide all non null lengths
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
       var type = shared.LengthValue.LengthType[i];
-      if (this[type] != null){
+      if (this[type] != null) {
         calcDictionary[type] = this[type] / divider;
       }
     }
@@ -93,14 +93,14 @@
   CalcLength.prototype._addCalcLengths = function(addedLength) {
     if (!(addedLength instanceof CalcLength)) {
       throw new TypeError('Objects not of type CalcLength');
-    } 
-    
+    }
+
     var calcDictionary = {};
 
     // Iterate through all possible length types and add their values
-    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
       var type = shared.LengthValue.LengthType[i];
-      if (this[type] == null){
+      if (this[type] == null) {
         calcDictionary[type] = addedLength[type];
       } else if (addedLength[type] == null) {
         calcDictionary[type] = this[type];
@@ -115,16 +115,16 @@
   CalcLength.prototype._subtractCalcLengths = function(subtractedLength) {
     if (!(subtractedLength instanceof CalcLength)) {
       throw new TypeError('Objects not of type CalcLength');
-    } 
-    
+    }
+
     var calcDictionary = {};
 
     // Iterate through all possible length types and add their values
-    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
       var type = shared.LengthValue.LengthType[i];
       if (subtractedLength[type] == null) {
         calcDictionary[type] = this[type];
-      } else if (this[type] == null){
+      } else if (this[type] == null) {
         calcDictionary[type] = -subtractedLength[type];
       } else {
         calcDictionary[type] = this[type] - subtractedLength[type];
@@ -139,14 +139,14 @@
   };
 
   CalcLength.prototype.equals = function(other) {
-    if (!(other instanceof CalcLength)){
+    if (!(other instanceof CalcLength)) {
       return false;
     }
 
     // Iterate through all length types and check that both objects contain the same values
-    for (var index in shared.LengthValue.LengthType){
+    for (var index in shared.LengthValue.LengthType) {
       var type = shared.LengthValue.LengthType[index];
-      if (this[type] != other[type]){
+      if (this[type] != other[type]) {
         return false;
       }
     }
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -138,15 +138,15 @@
     return this;
   };
 
-  CalcLength.prototype.equals = function(lengthValue) {
-    if (!(lengthValue instanceof CalcLength)){
+  CalcLength.prototype.equals = function(other) {
+    if (!(other instanceof CalcLength)){
       return false;
     }
 
     // Iterate through all length types and check that both objects contain the same values
     for (var index in shared.LengthValue.LengthType){
       var type = shared.LengthValue.LengthType[index];
-      if (this[type] != lengthValue[type]){
+      if (this[type] != other[type]){
         return false;
       }
     }
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -112,6 +112,30 @@
     return new CalcLength(calcDictionary);
   };
 
+  CalcLength.prototype._subtractCalcLengths = function(subtractedLength) {
+    if (!(subtractedLength instanceof CalcLength)) {
+      throw new TypeError('Objects not of type Calclength');
+    } 
+    
+    var calcDictionary = {};
+
+    //Iterate through all posible length types and add there values
+    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+      var type = shared.LengthValue.LengthType[i];
+      if (this[type] == null && subtractedLength[type] == null) {
+        calcDictionary[type] = null;
+      } else if (this[type] == null){
+        calcDictionary[type] = -subtractedLength[type];
+      } else if (subtractedLength[type] == null) {
+        calcDictionary[type] = -this[type];
+      } else {
+        calcDictionary[type] = this[type] - subtractedLength[type];
+      }
+    }
+
+    return new CalcLength(calcDictionary);
+  };
+
   LengthValue.prototype._convertToCalcLength = function() {
     return this;
   };
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -134,7 +134,7 @@
     return new CalcLength(calcDictionary);
   };
 
-  LengthValue.prototype._convertToCalcLength = function() {
+  LengthValue.prototype._asCalcLength = function() {
     return this;
   };
 
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -90,6 +90,28 @@
     return new CalcLength(calcDictionary);
   };
 
+  CalcLength.prototype._addCalcLengths = function(addedLength) {
+    if (!(addedLength instanceof CalcLength)) {
+      throw new TypeError('Objects not of type Calclength');
+    } 
+    
+    var calcDictionary = {};
+
+    //Iterate through all posible length types and add there values
+    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+      var type = shared.LengthValue.LengthType[i];
+      if (this[type] == null){
+        calcDictionary[type] = addedLength[type];
+      } else if (addedLength[type] == null) {
+        calcDictionary[type] = this[type];
+      } else {
+        calcDictionary[type] = this[type] + addedLength[type];
+      }
+    }
+
+    return new CalcLength(calcDictionary);
+  };
+
   LengthValue.prototype._convertToCalcLength = function() {
     return this;
   };
@@ -106,7 +128,7 @@
         return false;
       }
     }
-    
+
     return true;
   };
 
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -50,6 +50,13 @@
   LengthValue.prototype.add = function(addedLength) {
     if (this instanceof SimpleLength && addedLength instanceof SimpleLength && this.type == addedLength.type) {
       return this._addSimpleLengths(addedLength);
+
+    } else {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._convertToCalcLength();
+      var lengthToAdd_2 = addedLength._convertToCalcLength();
+
+      return lengthToAdd_1._addCalcLengths(lengthToAdd_2);
     }
     throw new TypeError('Not implemented yet');
   };
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -52,10 +52,7 @@
       return this._addSimpleLengths(addedLength);
     } else if (addedLength instanceof LengthValue) {
       //ensure both lengths are of type CalcLength before adding
-      var lengthToAdd_1 = this._asCalcLength();
-      var lengthToAdd_2 = addedLength._asCalcLength();
-
-      return lengthToAdd_1._addCalcLengths(lengthToAdd_2);
+      return this._asCalcLength()._addCalcLengths(addedLength._asCalcLength());
     } else {
       throw new TypeError('The object you are adding is not of type LengthValue');
     }
@@ -66,10 +63,7 @@
       return this._subtractSimpleLengths(subtractedLength);
     } else if (subtractedLength instanceof LengthValue) {
       //ensure both lengths are of type CalcLength before adding
-      var lengthToAdd_1 = this._asCalcLength();
-      var lengthToAdd_2 = subtractedLength._asCalcLength();
-
-      return lengthToAdd_1._subtractCalcLengths(lengthToAdd_2);
+      return this._asCalcLength()._subtractCalcLengths(subtractedLength._asCalcLength());
     } else {
       throw new TypeError('The object you are trying to subtract from this length is not of type LengthValue');
     }
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -64,6 +64,12 @@
   LengthValue.prototype.subtract = function(subtractedLength) {
     if (this instanceof SimpleLength && subtractedLength instanceof SimpleLength && this.type == subtractedLength.type) {
       return this._subtractSimpleLengths(subtractedLength);
+    } else {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._convertToCalcLength();
+      var lengthToAdd_2 = subtractedLength._convertToCalcLength();
+
+      return lengthToAdd_1._subtractCalcLengths(lengthToAdd_2);
     }
     throw new TypeError('Not implemented yet');
   };
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -50,28 +50,29 @@
   LengthValue.prototype.add = function(addedLength) {
     if (this instanceof SimpleLength && addedLength instanceof SimpleLength && this.type == addedLength.type) {
       return this._addSimpleLengths(addedLength);
-
-    } else {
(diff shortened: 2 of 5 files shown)


src/calc-length.js
```diff
@@ -93,6 +93,72 @@
...
   };
 
+  CalcLength.prototype._addCalcLengths = function(addedLength) {
+    if (!(addedLength instanceof CalcLength)) {
+      throw new TypeError('Objects not of type CalcLength');
+    } 
+    
+    var calcDictionary = {};
+
+    //Iterate through all possible length types and add there values
```
wilddamon: Nitpicks: 
- Add single space after //
- s/there/their

Same below

src/calc-length.js
```diff
@@ -93,6 +93,72 @@
...
+    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
```
wilddamon: Nitpick: need space after "for", and space between the closing ) and opening {

for (var i = 0; ...) {

Same elsewhere in this file

src/calc-length.js
```diff
@@ -93,6 +93,72 @@
...
+
+  CalcLength.prototype.equals = function(lengthValue) {
+    if (!(lengthValue instanceof CalcLength)){
+      return false;
+    }
+
+    //iterate through all length types and check that both objects contain the same values
+    for(var index in shared.LengthValue.LengthType){
+      var type = shared.LengthValue.LengthType[index];
+      if(this[type] != lengthValue[type]){
```
wilddamon: Also need space after if

if (condition ...

src/calc-length.js
```diff
@@ -93,6 +93,72 @@
...
+      throw new TypeError('Objects not of type CalcLength');
+    } 
+    
+    var calcDictionary = {};
+
+    //Iterate through all possible length types and add there values
+    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+      var type = shared.LengthValue.LengthType[i];
+      if (this[type] == null && subtractedLength[type] == null) {
+        calcDictionary[type] = null;
```
wilddamon, line 128: In the constructor, you check if things are undefined, so it's fine to skip this case and end up with calcDictionary[type] being undefined. In this case, it's clearer to invert the conditions in the if statements:

if (subtractedLength[type] != null && this[type] != null) {
  calcDictionary[type] = this[type] - subtractedLength[type];
} else if (this[type] != null) {
  calcDictionary[type] = this[type];
} else if (subtractedLength[type] != null) {
  calcDictionary[type] = -subtractedLength[type];
}

src/calc-length.js
```diff
@@ -93,6 +93,72 @@
...
+    if (!(addedLength instanceof CalcLength)) {
+      throw new TypeError('Objects not of type CalcLength');
+    } 
+    
+    var calcDictionary = {};
+
+    //Iterate through all possible length types and add there values
+    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+      var type = shared.LengthValue.LengthType[i];
+      if (this[type] == null){
```
wilddamon: It's clearer to (find the things that are defined and add them) than it is to (find the things that are undefined and ad the other thing).

if (addedLength[type] != null && this[type] != null) {
calcDictionary[type] = this[type] + addedLength[type];
} else if (this[type] != null) {
calcDictionary[type] = this[type];
} else if (addedLength[type] != null) {
calcDictionary[type] = addedLength[type];
}

src/calc-length.js
```diff
@@ -93,6 +93,72 @@
...
+
+    //Iterate through all possible length types and add there values
+    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+      var type = shared.LengthValue.LengthType[i];
+      if (this[type] == null && subtractedLength[type] == null) {
+        calcDictionary[type] = null;
+      } else if (this[type] == null){
+        calcDictionary[type] = -subtractedLength[type];
+      } else if (subtractedLength[type] == null) {
+        calcDictionary[type] = -this[type];
```
wilddamon: This shouldn't be negative.

src/length-value.js
```diff
@@ -50,13 +50,25 @@
...
       return this._addSimpleLengths(addedLength);
-    } 
+    } else {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._convertToCalcLength();
+      var lengthToAdd_2 = addedLength._convertToCalcLength();
+
+      return lengthToAdd_1._addCalcLengths(lengthToAdd_2);
+    }
     throw new TypeError('Not implemented yet');
```
wilddamon: This line can't be reached any more and can be deleted.

src/simple-length.js
```diff
@@ -67,6 +67,25 @@
     return new SimpleLength((this.value - subtractedLength.value), this.type);
   };
 
+  SimpleLength.prototype._convertToCalcLength = function() {
```
wilddamon: I think it's more common to use _asCalcLength - plus it's shorter! :)

src/length-value.js
```diff
@@ -50,13 +50,25 @@
   LengthValue.prototype.add = function(addedLength) {
     if (this instanceof SimpleLength && addedLength instanceof SimpleLength && this.type == addedLength.type) {
       return this._addSimpleLengths(addedLength);
-    } 
+    } else {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._convertToCalcLength();
+      var lengthToAdd_2 = addedLength._convertToCalcLength();
+
+      return lengthToAdd_1._addCalcLengths(lengthToAdd_2);
```
wilddamon: If you change _convertToCalcLength to _asCalcLength, it's fine to just write

return (this._asCalcLength())._addCalcLengths(addedLength._asCalcLength());

Similar below for subtract.

src/length-value.js
```diff
@@ -73,6 +85,12 @@
     throw new TypeError('Not implemented yet');
   };
 
+  LengthValue.prototype._convertToCalcLength = function() {
```
wilddamon: This should probably throw an error saying it shouldn't be reached.

src/length-value.js
```diff
@@ -73,6 +85,12 @@
...
+  };
+
+  LengthValue.prototype.equals = function(lengthValue) {
```
wilddamon: Same here, should probably never be reached.

src/simple-length.js
```diff
@@ -67,6 +67,25 @@
...
+  }
+
+  SimpleLength.prototype.equals = function(lengthValue) {
+    if (!(lengthValue instanceof SimpleLength)){
+      return false;
+    }
+    if (!(this.type == lengthValue.type && this.value == lengthValue.value)){
+      return false;
+    }
+      return true;
```
wilddamon: Fix indenting

src/simple-length.js
```diff
@@ -67,6 +67,25 @@
...
+    var calcDictionary = {};
+
+    //Set calcDictionary field to the Length type inside the SimpleLength object that called the function
+    calcDictionary[this.type] = this.value;
+
+    return new CalcLength(calcDictionary);
+  }
+
+  SimpleLength.prototype.equals = function(lengthValue) {
+    if (!(lengthValue instanceof SimpleLength)){
```
wilddamon: Needs space before {. Same below.

src/simple-length.js
```diff
@@ -67,6 +67,25 @@
...
+      return false;
+    }
+    if (!(this.type == lengthValue.type && this.value == lengthValue.value)){
```
wilddamon: It's shorter and clearer to write

if (this.type != lengthValue.type || this.value != lengthValue.value)

src/simple-length.js
```diff
@@ -67,6 +67,25 @@
...
+  SimpleLength.prototype._convertToCalcLength = function() {
+    var calcDictionary = {};
+
+    //Set calcDictionary field to the Length type inside the SimpleLength object that called the function
+    calcDictionary[this.type] = this.value;
+
+    return new CalcLength(calcDictionary);
+  }
+
+  SimpleLength.prototype.equals = function(lengthValue) {
```
wilddamon: It's more common to call the argument to .equals "other":

SimpleLength.prototype.equals = function(other) {

Same in the other files

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
     var dividedCalcLength = calcLength.divide(2.5);
     assert.strictEqual(dividedCalcLength.cssString, 'calc(10px+1.28em)');
   });
+
+  test('Adding two CalcLengths returns a new CalcLength with expected values in each value type', function() {
+    var cLength_1 = new CalcLength({px: 15, em: 6})
```
wilddamon: shouldn't contract names, so variables should be

var calcLength1
var calcLength2

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+    var cLength_2 = new CalcLength({px: 10, em: 3})
+    var lengthAddition = cLength_1.add(cLength_2);
+    var expectedlengthAddition = new CalcLength({px: 25, em: 9})
```
wilddamon: just call this "expected". Same elsewhere where you make a variable that stores an expected result. (You can also use "expectedResult" if you think that's clearer.

This is also a bit nitpicky, but it's often clearer to define all the necessary variables first, then construct the expectation, then do the function call under test, and finally do the assertions. So this function would look like:

{
var calcLength1 = new CalcLength({px: 15, em: 6});
var calcLength2 = new CalcLength({px: 10, em: 3});
var expectedResult = new CalcLength({px: 25, em: 9});

var result = calcLength1.add(calcLength2);

assert.isTrue(result.equals(expectedResult));
}

Also, you can reasonably assume that .equals returns false if the result and expected result have different types, so you can delete the line asserting that the result is a CalcLength.

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
     var dividedCalcLength = calcLength.divide(2.5);
     assert.strictEqual(dividedCalcLength.cssString, 'calc(10px+1.28em)');
   });
+
+  test('Adding two CalcLengths returns a new CalcLength with expected values in each value type', function() {
+    var cLength_1 = new CalcLength({px: 15, em: 6})
+    var cLength_2 = new CalcLength({px: 10, em: 3})
+    var lengthAddition = cLength_1.add(cLength_2);
```
wilddamon: just call this "result". Same elsewhere for variables that store the result for the test.

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+    var expectedlengthAddition = new CalcLength({px: 25, em: 9})
+
+    assert.instanceOf(lengthAddition, CalcLength, 'two added CalcLength of same type should be an instance of CalcLength');
+    assert.isTrue(expectedlengthAddition.equals(lengthAddition));
+  });
+
+    test('CalcLength do not have to have the same value types in order to be added and all value types seen in both'
+        +'CalcLengths with be in the outputted CalcLength', function() {
```
wilddamon: This is kind of long, how about

"All values in added CalcLengths appear in the output"

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+    var cLength_1 = new CalcLength({px: 15, em: 6})
+    var cLength_2 = new CalcLength({px: 10, em: 3})
+    var lengthSubtraction = cLength_1.subtract(cLength_2);
+    var expectedlengthSubtraction = new CalcLength({px: 5, em: 3})
+
+    assert.instanceOf(lengthSubtraction, CalcLength, 'two added CalcLength of same type should be an instance of CalcLength');
+    assert.isTrue(expectedlengthSubtraction.equals(lengthSubtraction));
+  });
+
+    test('CalcLength do not have to have the same value types in order to be subtracted from each other and all value types '
```
wilddamon: Likewise, how about

"All values in subtracted CalcLengths appear in the output"

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+    assert.isTrue(expectedlengthSubtraction.equals(lengthSubtraction));
+  });
+
+  test('convertToCalcLength method returns the object that called it if it is of type CalcLength', function() {
+    var cLength = new CalcLength({px: 25, em: 3.2})
+    var convertedCalcLength = cLength._convertToCalcLength();
+    assert.strictEqual(cLength, convertedCalcLength);
+  });
+
+  test('Test equals method of CalcLength such that it will return true if two CalcLength\'s are the same', function() {
```
wilddamon: Don't need the apostrophe.

How about:
'Equals method should return true for equal CalcLengths'

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+      +'even if all other types are equivalent', function() {
+  var cLength_1 = new CalcLength({px: 25, em: 3.2, percent: 5})
+  var cLength_2 = new CalcLength({px: 25, em: 3.2})
+  
+  assert.isFalse(cLength_1.equals(cLength_2));
+  });
+
+  test('Equals method should return true when one CalcLength has additional length types with null values', function() {
+  var cLength_1 = new CalcLength({px: 25, em: 3.2, percent: null})
+  var cLength_2 = new CalcLength({px: 25, em: 3.2})
```
wilddamon: These are represented the same internally so this test doesn't do anything and can be removed.

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+  });
+
+  test('Test equals method of CalcLength such that it will return true if two CalcLength\'s are the same', function() {
+  var cLength_1 = new CalcLength({px: 25, em: 3.2})
+  var cLength_2 = new CalcLength({px: 25, em: 3.2})
+  
+  assert.isTrue(cLength_1.equals(cLength_2));
+  });
+
+  test('Equals method should return false when one CalcLength has additional length types with non null values'
```
wilddamon: How about
"Equals method should return false for non-equal CalcLengths"?

test/js/simple-length.js
```diff
@@ -90,4 +90,24 @@ suite('SimpleLength', function() {
     assert.strictEqual(lengthAddition.value, 5);
   });
 
+  test('convertToCalcLength method returns a CalcLength object with a single length value type when called from a SimpleLength object', function() {
```
wilddamon: We already know it's being called from a SimpleLength object since we're in the SimpleLength test suite. How about

"asCalcLength method returns a CalcLength with single value"?

test/js/simple-length.js
```diff
@@ -90,4 +90,24 @@ suite('SimpleLength', function() {
...
+    var simpleLength = new SimpleLength(10, 'em');
+    var convertedCalcLength = simpleLength._convertToCalcLength();
+    var testCalcLength = new CalcLength({em: 10})
+    assert.isTrue(convertedCalcLength.equals(testCalcLength));
+  });
+
+  test('Test equals method of Simple length such that it will return true if two simple lengths are the same', function() {
```
wilddamon: Similar to the calc length tests, how about

"Equals method should return true for equal SimpleLengths"

test/js/simple-length.js
```diff
@@ -90,4 +90,24 @@ suite('SimpleLength', function() {
...
+  
+  assert.isTrue(simpleLength_1.equals(simpleLength_2));
+  });
+
+  test('Equals method should return false if a CalcLength is compared to a simple length even if they have the same value', function() {
+  var simpleLength_1 = new SimpleLength(10, 'em');
+  var simpleLength_2 = new CalcLength({em: 10});
+  
+  assert.isFalse(simpleLength_1.equals(simpleLength_2));
+  });
```
wilddamon, line 128: Please add a test for  "Equals method should return false for non-equal SimpleLengths" that checks .equals returns false if the value or the type is different.

test/js/simple-length.js
```diff
@@ -90,4 +90,24 @@ suite('SimpleLength', function() {
...
+
+  test('Test equals method of Simple length such that it will return true if two simple lengths are the same', function() {
+  var simpleLength_1 = new SimpleLength(10, 'em');
+  var simpleLength_2 = new SimpleLength(10, 'em');
+  
+  assert.isTrue(simpleLength_1.equals(simpleLength_2));
+  });
+
+  test('Equals method should return false if a CalcLength is compared to a simple length even if they have the same value', function() {
```
wilddamon: How about "Equals method should return false if any other kind of object is passed in"

then you can also try

assert.isFalse(simpleLength.equals({}));
assert.isFalse(simpleLength.equals(5));

test/js/simple-length.js
```diff
@@ -90,4 +90,24 @@ suite('SimpleLength', function() {
...
+  var simpleLength_1 = new SimpleLength(10, 'em');
+  var simpleLength_2 = new CalcLength({em: 10});
```
wilddamon: this isn't a SimpleLength.

test/js/simple-length.js
```diff
@@ -90,4 +90,24 @@ suite('SimpleLength', function() {
     assert.strictEqual(lengthAddition.value, 5);
   });
 
+  test('convertToCalcLength method returns a CalcLength object with a single length value type when called from a SimpleLength object', function() {
+    var simpleLength = new SimpleLength(10, 'em');
+    var convertedCalcLength = simpleLength._convertToCalcLength();
+    var testCalcLength = new CalcLength({em: 10})
```
wilddamon: Call this "expected"

test/js/simple-length.js
```diff
@@ -90,4 +90,24 @@ suite('SimpleLength', function() {
     assert.strictEqual(lengthAddition.value, 5);
   });
 
+  test('convertToCalcLength method returns a CalcLength object with a single length value type when called from a SimpleLength object', function() {
+    var simpleLength = new SimpleLength(10, 'em');
+    var convertedCalcLength = simpleLength._convertToCalcLength();
```
wilddamon: Call this "result"

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+  });
+
+    test('CalcLength do not have to have the same value types in order to be subtracted from each other and all value types '
+        +'seen in both CalcLengths with be in the outputted CalcLength', function() {
+    var cLength_1 = new CalcLength({px: 15, em: 6, percent: 5})
+    var cLength_2 = new CalcLength({px: 10, em: 3, ex: 6})
+    var lengthSubtraction = cLength_1.subtract(cLength_2);
+    var expectedlengthSubtraction = new CalcLength({px: 5, em: 3, percent: -5, ex: -6});
+
+    assert.instanceOf(lengthSubtraction, CalcLength, 'two added CalcLength of same type should be an instance of CalcLength');
```
wilddamon: Kind of verbose too. How something like "Result from adding two CalcLengths should be a CalcLength"?

test/js/calc-length.js
```diff
@@ -83,4 +83,74 @@ suite('CalcLength', function() {
...
+  
+  assert.isFalse(cLength_1.equals(cLength_2));
+  });
+
+  test('Equals method should return true when one CalcLength has additional length types with null values', function() {
+  var cLength_1 = new CalcLength({px: 25, em: 3.2, percent: null})
+  var cLength_2 = new CalcLength({px: 25, em: 3.2})
+  
+  assert.isTrue(cLength_1.equals(cLength_2));
+  });
```
wilddamon, line 170: It would also be good to check that .equals returns false if the passed in object is not a CalcLength.

src/simple-length.js
```diff
@@ -67,6 +67,25 @@
     return new SimpleLength((this.value - subtractedLength.value), this.type);
   };
 
+  SimpleLength.prototype._convertToCalcLength = function() {
+    var calcDictionary = {};
+
+    //Set calcDictionary field to the Length type inside the SimpleLength object that called the function
```
wilddamon: This comment is unnecessary. Please delete it.

src/calc-length.js
```diff
@@ -93,6 +93,70 @@
...
+    return this;
+  };
+
+  CalcLength.prototype.equals = function(other) {
+    if (!(other instanceof CalcLength)){
+      return false;
+    }
+
+    // Iterate through all length types and check that both objects contain the same values
+    for (var index in shared.LengthValue.LengthType){
```
rjwright: 'index' should just be 'i'. Same elsewhere in this file. Loop counters can be given special names when there's a good reason to do so; otherwise just use i, j, k.

src/calc-length.js
```diff
@@ -69,9 +69,9 @@
     var calcDictionary = {};
 
     // Iterate through all length types and multiply all non null lengths 
-    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
       var type = shared.LengthValue.LengthType[i];
-      if(this[type] != null){
+      if (this[type] != null){
```
rjwright: Missing space between bracket and brace here. Also on line 86, 88, 104, 106, 126, 130, 145, 150, 152.

src/calc-length.js
```diff
@@ -69,9 +69,9 @@
     var calcDictionary = {};
 
     // Iterate through all length types and multiply all non null lengths 
-    for(var i = 0; i < shared.LengthValue.LengthType.length; i++){
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++){
```
rjwright: Missing space between bracket and brace here. Also on line 74, 86, 88, 104, 106, 126, 130, 145, 150, 152.

src/calc-length.js
```diff
@@ -93,6 +93,70 @@
     return new CalcLength(calcDictionary);
   };
 
+  CalcLength.prototype._addCalcLengths = function(addedLength) {
+    if (!(addedLength instanceof CalcLength)) {
+      throw new TypeError('Objects not of type CalcLength');
```
rjwright: Maybe make the message 'Argument must be a CalcLength'. I assume this was copied from SimpleLength; feel free to also change it there.

src/length-value.js
```diff
@@ -50,15 +50,29 @@
...
     if (this instanceof SimpleLength && addedLength instanceof SimpleLength && this.type == addedLength.type) {
       return this._addSimpleLengths(addedLength);
-    } 
-    throw new TypeError('Not implemented yet');
+    } else if (addedLength instanceof LengthValue) {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._asCalcLength();
+      var lengthToAdd_2 = addedLength._asCalcLength();
+
+      return lengthToAdd_1._addCalcLengths(lengthToAdd_2);
```
rjwright: I think Eddy might have mentioned this in a previous review. You can just say

```
return this._asCalcLength()._addCalcLengths(addedLength._asCalcLength());
```

Also, for future reference, please don't use underscores in any identifier names except for object members that you wish to be treated as private. E.g. here you would use lengthToAdd1, lengthToAdd2.

src/length-value.js
```diff
@@ -50,15 +50,29 @@
...
 
   LengthValue.prototype.subtract = function(subtractedLength) {
     if (this instanceof SimpleLength && subtractedLength instanceof SimpleLength && this.type == subtractedLength.type) {
       return this._subtractSimpleLengths(subtractedLength);
+    } else if (subtractedLength instanceof LengthValue) {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._asCalcLength();
+      var lengthToAdd_2 = subtractedLength._asCalcLength();
+
+      return lengthToAdd_1._subtractCalcLengths(lengthToAdd_2);
```
rjwright: Again, no need to make lengthToAdd\* variables.

src/length-value.js
```diff
@@ -50,15 +50,29 @@
...
-    } 
-    throw new TypeError('Not implemented yet');
+    } else if (addedLength instanceof LengthValue) {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._asCalcLength();
+      var lengthToAdd_2 = addedLength._asCalcLength();
+
+      return lengthToAdd_1._addCalcLengths(lengthToAdd_2);
+    } else {
+      throw new TypeError('The object you are adding is not of type LengthValue');
```
rjwright: Make this error message consistent with the ones in calc-length.js

src/length-value.js
```diff
@@ -50,15 +50,29 @@
...
     if (this instanceof SimpleLength && subtractedLength instanceof SimpleLength && this.type == subtractedLength.type) {
       return this._subtractSimpleLengths(subtractedLength);
+    } else if (subtractedLength instanceof LengthValue) {
+      //ensure both lengths are of type CalcLength before adding
+      var lengthToAdd_1 = this._asCalcLength();
+      var lengthToAdd_2 = subtractedLength._asCalcLength();
+
+      return lengthToAdd_1._subtractCalcLengths(lengthToAdd_2);
+    } else {
+      throw new TypeError('The object you are trying to subtract from this length is not of type LengthValue');
```
rjwright: Make this error message consistent with the ones in calc-length.js

src/simple-length.js
```diff
@@ -67,6 +67,25 @@
     return new SimpleLength((this.value - subtractedLength.value), this.type);
   };
 
+  SimpleLength.prototype._asCalcLength = function() {
+    var calcDictionary = {};
+
```
rjwright: These blank lines are unnecessary. I think it's good to use whitespace to break up blocks (exactly when is a matter of taste), but for very small blocks you should just bunch them up.

src/calc-length.js
```diff
@@ -90,6 +90,70 @@
...
+      throw new TypeError('Argument must be a CalcLength');
+    }
+
+    var calcDictionary = {};
+
+    // Iterate through all possible length types and add their values
+    for (var i = 0; i < shared.LengthValue.LengthType.length; i++) {
+      var type = shared.LengthValue.LengthType[i];
+      if (this[type] == null) {
+        calcDictionary[type] = addedLength[type];
```
wilddamon, line 106: You need to check if both are defined before you do this test case. If you do it this way around, addedLength[type] can still be null here. This sort of works, because
1 + null = null
null + 2 = 2

however
null + null = 0

So you will potentially end up with extra zeros in your output calclength.

This is why it's easier to check if things are defined rather than whether they are undefined.

Same below.

src/length-value.js
```diff
@@ -50,15 +50,23 @@
   LengthValue.prototype.add = function(addedLength) {
     if (this instanceof SimpleLength && addedLength instanceof SimpleLength && this.type == addedLength.type) {
       return this._addSimpleLengths(addedLength);
+    } else if (addedLength instanceof LengthValue) {
+      //ensure both lengths are of type CalcLength before adding
```
wilddamon: Nitpick: Comment should start with capital letter, need space after //. Same for subtract.

src/length-value.js
```diff
@@ -73,6 +81,12 @@
     throw new TypeError('Not implemented yet');
   };
 
+  LengthValue.prototype._asCalcLength = function() {
```
wilddamon, line 84: throw new TypeError("Should not be reached"). Same for equals

src/simple-length.js
```diff
@@ -64,6 +64,23 @@
     return new SimpleLength((this.value - subtractedLength.value), this.type);
   };
 
+  SimpleLength.prototype._asCalcLength = function() {
+    var calcDictionary = {};
+    //Set calcDictionary field to the Length type inside the SimpleLength object that called the function
```
wilddamon: Nitpick: Need space after //.

test/js/calc-length.js
```diff
@@ -61,26 +61,94 @@ suite('CalcLength', function() {
...
-    var calcLength = new CalcLength({px: 25, em: 3.2})
+    var calcLength = new CalcLength({px: 25, em: 3.2});
     var dividedCalcLength = calcLength.divide(2.5);
     assert.strictEqual(dividedCalcLength.cssString, 'calc(10px+1.28em)');
   });
+
+  test('Adding two CalcLengths returns a new CalcLength with expected values in each value type', function() {
+    var calcLength1 = new CalcLength({px: 15, em: 6});
+    var calcLength2 = new CalcLength({px: 10, em: 3});
+    var lengthAddition = calcLength1.add(calcLength2);
```
wilddamon: Just call this "result". Same for similar variables elsewhere. It makes it clearer that this is the result of an operation.

test/js/calc-length.js
```diff
@@ -61,26 +61,94 @@ suite('CalcLength', function() {
...
+
+    assert.instanceOf(lengthAddition, CalcLength, 'two added CalcLength of same type should be an instance of CalcLength');
+    assert.isTrue(expectedlengthAddition.equals(lengthAddition));
+  });
+
+    test('All values in added CalcLengths appear in the output', function() {
+    var calcLength1 = new CalcLength({px: 15, em: 6, percent: 5});
+    var calcLength2 = new CalcLength({px: 10, em: 3, ex: 6});
+    var lengthAddition = calcLength1.add(calcLength2);
+    var expectedlengthAddition = new CalcLength({px: 25, em: 9, percent: 5, ex: 6});
```
wilddamon: result and expectedResult. It's more common so it's easier to read at a glance.

src/calc-length.js
```diff
@@ -90,6 +90,74 @@
...
+    return this;
+  };
+
+  CalcLength.prototype.equals = function(other) {
+    if (!(other instanceof CalcLength)) {
+      return false;
+    }
+
+    // Iterate through all length types and check that both objects contain the same values
+    for (var index in shared.LengthValue.LengthType) {
```
rjwright: 'index' should just be 'i'. Same elsewhere in this file. Loop counters can be given special names when there's a good reason to do so; otherwise just use i, j, k.

src/length-value.js
```diff
@@ -50,27 +50,39 @@
   LengthValue.prototype.add = function(addedLength) {
     if (this instanceof SimpleLength && addedLength instanceof SimpleLength && this.type == addedLength.type) {
       return this._addSimpleLengths(addedLength);
+    } else if (addedLength instanceof LengthValue) {
+      // Ensure both lengths are of type CalcLength before adding
+      return this._asCalcLength()._addCalcLengths(addedLength._asCalcLength());
+    } else {
+      throw new TypeError('The object you are adding is not of type LengthValue');
```
rjwright: Make this error message consistent with the ones in calc-length.js

src/length-value.js
```diff
@@ -50,27 +50,39 @@
...
   };
 
   LengthValue.prototype.subtract = function(subtractedLength) {
     if (this instanceof SimpleLength && subtractedLength instanceof SimpleLength && this.type == subtractedLength.type) {
       return this._subtractSimpleLengths(subtractedLength);
+    } else if (subtractedLength instanceof LengthValue) {
+      // Ensure both lengths are of type CalcLength before adding
+      return this._asCalcLength()._subtractCalcLengths(subtractedLength._asCalcLength());
+    } else {
+      throw new TypeError('The object you are trying to subtract from this length is not of type LengthValue');
```
rjwright: Make this error message consistent with the ones in calc-length.js

test/js/simple-length.js
```diff
@@ -46,48 +46,66 @@ suite('SimpleLength', function() {
...
 
+  test('asCalcLength method returns a CalcLength with single value', function() {
+    var simpleLength = new SimpleLength(10, 'em');
+    var result = simpleLength._asCalcLength();
+    var expectedResult = new CalcLength({em: 10});
+    assert.isTrue(result.equals(expectedResult));
+  });
+
+  test('Equals method should return true for equal SimpleLengths', function() {
+  var simpleLength1 = new SimpleLength(10, 'em');
```
rjwright: Indenting. Same with next testcase.

src/simple-length.js
```diff
@@ -64,6 +64,23 @@
     return new SimpleLength((this.value - subtractedLength.value), this.type);
   };
 
+  SimpleLength.prototype._asCalcLength = function() {
+    var calcDictionary = {};
+    // Set calcDictionary field to the Length type inside the SimpleLength object that called the function
```
rjwright: Unnecessary comment.

src/length-value.js
```diff
@@ -50,27 +50,39 @@
...
   LengthValue.prototype.divide = function(divider) {
-    throw new TypeError('Not implemented yet');
   };
 
   LengthValue.prototype.parse = function(cssString) {
-    throw new TypeError('Not implemented yet');
+    throw new TypeError('Should not be reached');
+  };
+
+  LengthValue.prototype._asCalcLength = function() {
```
rjwright: As previously requested by Eddy: "throw new TypeError("Should not be reached"). Same for equals"

test/js/calc-length.js
```diff
@@ -61,26 +61,114 @@ suite('CalcLength', function() {
...
+    assert.isTrue(expectedResult.equals(result));
+  });
+
+  test('All values in added CalcLengths appear in the output', function() {
+    var calcLength1 = new CalcLength({px: 15, em: 6, percent: 5});
+    var calcLength2 = new CalcLength({px: 10, em: 3, ex: 6});
+    var result = calcLength1.add(calcLength2);
+    var expectedResult = new CalcLength({px: 25, em: 9, percent: 5, ex: 6});
+
+    assert.instanceOf(result, CalcLength, 'two added CalcLength of same type should be an instance of CalcLength');
```
rjwright: Not sure what you mean by "of the same type". Clearly they're both CalcLengths, but they hold differently shaped data.

Also s/CalcLength/CalcLengths. Same throughout file.

test/js/calc-length.js
```diff
@@ -61,26 +61,114 @@ suite('CalcLength', function() {
...
+    assert.isTrue(expectedResult.equals(result));
+  });
+
+  test('length values in CalcLengths set to null add to null not zero', function() {
```
rjwright: Maybe "Null length values in CalcLengths add to null, not zero".

test/js/calc-length.js
rjwright: And actually we don't really need all of these instanceOf asserts for add etc. They are repetitive, but also LengthValue.equals checks types. You can leave in one for each method (one for add, one for subtract, ect.) for the sake of explicitness if you want. The rest can go.

rjwright: And if you cut it down just one per method, change the error message to "[operation]ing two CalcLengths returns a CalcLength".

test/js/calc-length.js
```diff
@@ -61,26 +61,114 @@ suite('CalcLength', function() {
...
+    var calcLength1 = new CalcLength({px: 15, em: 6, percent: 5});
+    var calcLength2 = new CalcLength({px: 10, em: 3, ex: 6});
+    var result = calcLength1.subtract(calcLength2);
+    var expectedResult = new CalcLength({px: 5, em: 3, percent: 5, ex: -6});
+
+    assert.instanceOf(result, CalcLength, 'two added CalcLength of same type should be an instance of CalcLength');
+    assert.isTrue(expectedResult.equals(result));
+  });
+
+  test('length values in CalcLengths set to null when subtracted will equal null not zero', function() {
```
rjwright: "Subtracting null length values in CalcLengths results in null, not zero"?

test/js/calc-length.js
```diff
@@ -61,26 +61,114 @@ suite('CalcLength', function() {
...
+    assert.isTrue(expectedResult.equals(result));
+  });
+
+  test('convertToCalcLength method returns the object that called it if it is of type CalcLength', function() {
+    var calcLength = new CalcLength({px: 25, em: 3.2});
+    var result = calcLength._asCalcLength();
+    assert.strictEqual(calcLength, result);
+  });
+
+  test('Test equals method of CalcLength such that it will return true if two CalcLength\'s are the same', function() {
```
rjwright: "CalcLength.equals returns true if the compared CalcLengths are the same."

test/js/calc-length.js
```diff
@@ -61,26 +61,114 @@ suite('CalcLength', function() {
...
+  var calcLength1 = new CalcLength({px: 25, em: 3.2});
```
rjwright: Indenting. And below.

test/js/simple-length.js
```diff
@@ -46,48 +46,66 @@ suite('SimpleLength', function() {
...
-    var length1 = new SimpleLength(10, 'em');
-    var length2 = new SimpleLength(5, 'em');
-    var lengthAddition = length1.add(length2);
-    assert.instanceOf(lengthAddition, SimpleLength, 'two added SimpleLength of same type should be an instance of SimpleLength');
-    assert.strictEqual(lengthAddition.type, 'em');
-    assert.strictEqual(lengthAddition.value, 15);
+    var simpleLength1 = new SimpleLength(10, 'em');
+    var simpleLength2 = new SimpleLength(5, 'em');
+    var result = simpleLength1.add(simpleLength2);
+    assert.instanceOf(result, SimpleLength, 'two added SimpleLength of same type should be an instance of SimpleLength');
```
rjwright: s/SimpleLength/SimpleLengths. And below.

rjwright:
LGTM when comments on test files are resolved. Will wait for LGTM from @wilddamon  before landing.

src/calc-length.js
```diff
@@ -90,6 +90,74 @@
...
+        calcDictionary[type] = -subtractedLength[type];
+      } else {
+        calcDictionary[type] = this[type] - subtractedLength[type];
+      }
+    }
+
+    return new CalcLength(calcDictionary);
+  };
+
+  LengthValue.prototype._asCalcLength = function() {
```
rjwright: This needs to be CalcLength.prototype._asCalcLength, right?

test/js/calc-length.js
```diff
@@ -61,26 +61,113 @@ suite('CalcLength', function() {
...
+    assert.isTrue(expectedResult.equals(result));
+  });
+
+  test('Null length values in CalcLengths add to null, not zero', function() {
+    var calcLength1 = new CalcLength({px: 15, em: null, percent: 5});
+    var calcLength2 = new CalcLength({px: 10, em: null, ex: 6});
+    var result = calcLength1.add(calcLength2);
+    var expectedResult = new CalcLength({px: 25, em: null, percent: 5, ex: 6});
+
+    assert.isTrue(expectedResult.equals(result));
```
wilddamon: In this case it is better to check that 

assert.strictEqual(null, result.em)

test/js/calc-length.js
```diff
@@ -61,26 +61,113 @@ suite('CalcLength', function() {
...
+  });
+
+  test('Subtracting null length values in CalcLengths results in null, not zero', function() {
+    var calcLength1 = new CalcLength({px: 15, em: null, percent: 5});
+    var calcLength2 = new CalcLength({px: 10, em: null, ex: 6});
+    var result = calcLength1.subtract(calcLength2);
+    var expectedResult = new CalcLength({px: 5, em: null, percent: 5, ex: -6});
+
+    assert.isTrue(expectedResult.equals(result));
```
wilddamon: Likewise here...

wilddamon:
Two very minor comments, otherwise LGTM

PR #29, opened 2015-12-03
Computed map properties

nattridgell:
- adds a list of properties that take length values
  to ComputedStylePropertyMap
  - adds a check for keywords
    NOTE: this may have to change / move around
  - adds some basic tests for length and keyword values
    for the map
    NOTE: These tests are currently being skipped as
     parsing is unimplemented.
  - adds `auto` keyword to KeywordValue
  - alter KeywordValue related tests as appropriate

### DIFF
--- a/src/computed-style-property-map.js
+++ b/src/computed-style-property-map.js
@@ -38,9 +38,9 @@
     if (!value) {
       return null;
     }
-    if (value == 'inherit') {
-      // TODO: Other keywords
-      throw new TypeError('Not implemented yet');
+
+    if (scope.KeywordValue.isKeywordValue(value)) {
+      return new scope.KeywordValue(value);
     }
 
     // TODO: The rest of the properties once the rest of the StyleValues are
@@ -55,6 +55,31 @@
       case 'widows':
       case 'z-index':
         return new scope.NumberValue(value);
+
+      // These properties take only a length value (length and percentage),
+      // or a keyword handled above.
+      case 'bottom':
+      case 'height':
+      case 'left':
+      case 'max-height':
+      case 'max-width':
+      case 'min-height':
+      case 'min-width':
+      case 'right':
+      case 'text-indent':
+      case 'top':
+      case 'width':
+        // TODO: ensure that LengthValue.parse is being implemented,
+        // or update to other appropriate function.
+        return new scope.LengthValue.parse(value);
+
+      // These properties only take a length value, but do not take percentage.
+      // TODO: Validation and implementation of this case.
+      case 'letter-spacing':
+      case 'word-spacing':
+        throw new TypeError('Not implemented yet');
+
+      // These properties take a number or SimpleLength
 
       case 'line-height':
         // normal | <number> | <length> | <percentage> | inherit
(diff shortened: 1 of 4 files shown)


nattridgell:
Now that we are temporarily returning to the div-soup approach for keywords, I have re-opened this PR, with the small change/comment addressed from the previous code review.

src/keyword-value.js
```diff
@@ -22,7 +22,7 @@
     this.cssString = value;
   }
 
-  KeywordValue.StyleValueKeyword = ['initial', 'inherit', 'revert', 'unset'];
+  KeywordValue.StyleValueKeyword = ['auto', 'initial', 'inherit', 'revert', 'unset'];
```
wilddamon, line 25: Perhaps this should be GlobalKeyword, and other things that refer to it for parsing/validation purposes should also refer to "isGlobalKeyword" and such. I think the keywords for particular types should go in the classes used for those purposes. It's probably still worth discussion of what the structure should look like for something like "speech-rate" which takes 

<number> | x-slow | slow | medium | fast | x-fast | faster | slower | inherit

since it doesn't look like there should be a SpeechRateValue class...

nattridgell, line 25: Cool beans :-)  I'll re-mock it up :D

rjwright, line 25: > It's probably still worth discussion of what the structure should look like for something like "speech-rate" which takes
> 
> | x-slow | slow | medium | fast | x-fast | faster | slower | inherit
> 
> since it doesn't look like there should be a SpeechRateValue class...

I think having a SpeechRateValue class just for the purpose of basing a  SpeechRateKeywordValue class would be OK. But there are many things you could do; we should discuss it.

rjwright, line 25: Wait, speech-rate can also take a number. That changes things.

src/computed-style-property-map.js
```diff
@@ -78,15 +103,28 @@
 
     switch (property) {
```
rjwright: This switch is getting large. Do we need it? ComputedStylePropertyMap.get throws for unsupported properties. Can we just do

```
var value = this.get(property);
if (Array.isArray(value)) {
    // TODO: Handle shorthands
} else {
    return value ? [value] : [];
}
```

here?

nattridgell: Sounds like a good idea for now, might need to restructure later on.  It is getting quite unwieldy, I agree.

src/keyword-value.js
nattridgell, line 25: As we are going for the keyword soup approach, I don't think "isGlobalKeyword" is as appropriate as before, nor do I think that the following comment is still applicable.  I stand to be corrected though.
"I think the keywords for particular types should go in the classes used for those purposes."

So, I'll just let this be for now, unless otherwise advised.

src/computed-style-property-map.js
```diff
@@ -56,6 +56,31 @@
...
+        // or update to other appropriate function.
+        return new scope.LengthValue.parse(value);
+
+      // These properties only take a length value, but do not take percentage.
+      // TODO: Validation and implementation of this case.
+      case 'letter-spacing':
+      case 'word-spacing':
+        throw new TypeError('Not implemented yet');
+
+      // These properties take a number or SimpleLength
```
nattridgell, line 82: This comment should be removed.

nattridgell:
Irrelevant with code restructure of computed style property map.

PR #31, opened 2015-12-04
Linter sweep

rjwright:
fix gjslint errors for

src/
- style-value
- number-value
- style-property-map
- computed-style-property-map
- scope

test/js/
- number-value
- keyword-value
- computed-style-property-map


nattridgell:
LGTM :+1: Thanks again for doing that.

wilddamon:
LGTM

PR #32, opened 2015-12-07
Error for when SimpleLength is constructed with an invalid type arg.

rjwright:
- Add error message for when SimpleLength is constructed with an invalid type arg
- Switch to not accepting numeric strings in NumberValue and SimpleLength constructors

### DIFF
--- a/src/simple-length.js
+++ b/src/simple-length.js
@@ -16,21 +16,14 @@
 
   // TODO: SimpleLength(simpleLength), SimpleLength(cssString)
   function SimpleLength(value, type) {
-    if (arguments.length == 2 && shared.LengthValue.LengthType.indexOf(type) >= 0) {
-      this.type = type;
-      if (typeof value == 'number') {
-        this.value = value;
-      } else if (typeof value == 'string') {
-        var nValue = Number.parseFloat(value);
-        if (!isNaN(nValue)) {
-          this.value = nValue;
-        }
-      }
+    if (typeof value != 'number') {
+      throw new TypeError('Value of SimpleLength must be a number.');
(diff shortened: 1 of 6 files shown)


src/simple-length.js
```diff
@@ -26,6 +26,8 @@
           this.value = nValue;
         }
       }
+    } else {
+      throw new TypeError('\'' + type + '\' is not a valid type for a SimpleLength.');
     }
     if (this.value == undefined) {
       throw new TypeError('Value of SimpleLength must be a number or a numeric string.');
```
wilddamon: While you're here, could you move this check up to the top? It's kind of weird to have it down here, especially once you have the check you've just added.

rjwright: To do that I'd need to re-check everything in the first if statement (i.e. that value is a number or a numeric string). I think it makes more sense here.

I've changed the error check in number-value.js to match this one.

wilddamon: I thought we were getting rid of the string constructors anyway, so it would end up being something more like

function SimpleLength(value, type) {
if (typeof value != 'number') {
throw "value must be a number"
}
if (shared.LengthValue.LengthType.indexOf(type) < 0) {
throw "invalid argument for unit"
}

this.value = value;
this.type = type;
}

Which I think is much more clear because the validation and the actual setting of the member values is grouped. The check for number of arguments is unnecessary, as extra arguments will be ignored, and if too few are specified, value and/or type will just be undefined.

rjwright: Good point :)

wilddamon:
LGTM

PR #34, opened 2015-12-09
StylePropertyMap restructure

njlawton:
Restructuring the current StylePropertyMap classes to be consistent with the original spec


src/style-property-map-readonly.js
```diff
@@ -14,27 +14,34 @@
...
+    get: function(property) {},
+    getAll: function(property) {},
+    has: function(property) {},
+    set: function(property, value) {},
+    getProperties: function() {}
+  };
 
-  ComputedStylePropertyMap.prototype.append = function(property, value) {
-    throw new TypeError('ComputedStylePropertyMap is immutable');
+  StylePropertyMapReadOnly.prototype.append = function(property, value) {
```
wilddamon: The modification functions should just not exist on StylePropertyMapReadOnly - they can just be added to StylePropertyMap (and the error can be "Not implemented yet" now instead).

test/js/inline-style-property-map.js
```diff
@@ -0,0 +1,16 @@
+suite('Inline StylePropertyMap', function() {
+  setup(function() {
+    this.element = document.createElement('div');
+    document.documentElement.appendChild(this.element);
```
wilddamon: It's more common to write

document.body.appendChild(element)

Otherwise the element ends up going after the body, which is a little bit strange.

test/js/inline-style-property-map.js
```diff
@@ -0,0 +1,16 @@
...
+    this.element.style.opacity = '0.5';
+  });
+  teardown(function() {
+    document.documentElement.removeChild(this.element);
+  });
+
+  test('Test that a new InlineStylePropertyMap is created when a HTML element referance calls the ' +
+	    'styleMap method', function() {
+	 var inlineStyleMap = this.element.styleMap();
```
wilddamon: Fix indenting, should only be 2 spaces in from test(.

src/style-property-map.js
```diff
@@ -14,16 +14,18 @@
...
 
-  StylePropertyMap.prototype = {
-    append: function(property, value) {},
-    delete: function(property) {},
-    get: function(property) {},
-    getAll: function(property) {},
-    has: function(property) {},
-    set: function(property, value) {},
-    getProperties: function() {}
+  StylePropertyMap.prototype.prototype = StylePropertyMapReadOnly.prototype;
```
wilddamon: Right now you're adding stuff to the ReadOnly one's prototype too. I have some ideas about how to fix the prototype chain which I'll look into as well, but in the meantime it's best to make a copy.

src/style-property-map-readonly.js
```diff
@@ -14,27 +14,26 @@
...
-    Object.create(shared.StylePropertyMap.prototype);
-
-  ComputedStylePropertyMap.prototype.append = function(property, value) {
-    throw new TypeError('ComputedStylePropertyMap is immutable');
-  };
-
-  ComputedStylePropertyMap.prototype.delete = function(property) {
-    throw new TypeError('ComputedStylePropertyMap is immutable');
+  StylePropertyMapReadOnly.prototype = {
+    append: function(property, value) {},
```
wilddamon: This wasn't quite what I meant. The StylePropertyMapReadOnly should not have any of the mutating methods. 
append, delete and set should all be moved into StylePropertyMap, and this map can be removed.

test/js/computed-style-property-map.js
```diff
@@ -8,16 +8,9 @@ suite('ComputedStylePropertyMap', function() {
...
-    var computedStyleMap = new ComputedStylePropertyMap(this.element);
-    assert.instanceOf(computedStyleMap, ComputedStylePropertyMap);
-    assert.instanceOf(computedStyleMap, typedOMTesting.StylePropertyMap);
-  });
-
-  test('ComputedStylePropertyMap.get works for properties that can only be ' +
+  test('Computed StylePropertyMap.get works for properties that can only be ' +
       'NumberValues', function() {
-    var computedStyleMap = new ComputedStylePropertyMap(this.element);
+    var computedStyleMap = new StylePropertyMapReadOnly(getComputedStyle(this.element));
```
wilddamon: Please use the new getComputedStyleMap function instead

test/js/inline-style-property-map.js
```diff
@@ -0,0 +1,16 @@
...
+  setup(function() {
+    this.element = document.createElement('div');
+    document.body.appendChild(this.element);
+    this.element.style.opacity = '0.5';
+  });
+  teardown(function() {
+    document.body.removeChild(this.element);
+  });
+
+  test('Test that a new InlineStylePropertyMap is created when a HTML element referance calls the ' +
```
wilddamon: This test seems suspicious - there is no InlineStylePropertyMap anymore? Perhaps it should be "Test that element.styleMap returns an Inline StylePropertyMap"

Also the formatting in this function is still weird...

njlawton:
Should be good now

src/style-property-map-readonly.js
```diff
@@ -103,27 +92,26 @@
 
   };
 
-  ComputedStylePropertyMap.prototype.set = function(property, value) {
-    throw new TypeError('ComputedStylePropertyMap is immutable');
+  StylePropertyMapReadOnly.prototype.set = function(property, value) {
```
wilddamon: You can delete set now too.

rjwright:
So just to be clear, the purpose of this CL is to
- remove ComputedStylePropertyMap and move its logic into StylePropertyMapReadOnly
- make StylePropertyMap instantiable and use it as the return type for Element.styleMap (the inline style map)
- make StylePropertyMap extend StylePropertyMapReadOnly
- make the StylePropertyMap/StylePropertyMapReadOnly constructors take a style object as an argument

right?

test/js/inline-style-property-map.js
```diff
@@ -0,0 +1,15 @@
...
+  setup(function() {
+    this.element = document.createElement('div');
+    document.body.appendChild(this.element);
+    this.element.style.opacity = '0.5';
+  });
+  teardown(function() {
+    document.body.removeChild(this.element);
+  });
+
+  test('Test that a new InlineStylePropertyMap is created when a HTML element referance calls the styleMap method', function() {
```
rjwright: s/InlineStylePropertyMap/StylePropertyMap

I would change this to "Element.styleMap returns a StylePropertyMap."

test/js/inline-style-property-map.js
```diff
@@ -0,0 +1,15 @@
...
+    var inlineStyleMap = this.element.styleMap();
+    assert.instanceOf(inlineStyleMap, StylePropertyMap, 'The styleMap method should return an instance of InlineStylePropertyMap');
```
rjwright: s/InlineStylePropertyMap/StylePropertyMap

src/style-property-map.js
rjwright: This is what I told him to do. But it was actually wrong, it should have been 

```
StylePropertyMap.prototype.__proto__ = StylePropertyMapReadOnly.prototype;
```

Sorry Nick.

But using the inherit method is neater anyway.

test/js/computed-style-property-map.js
```diff
@@ -8,16 +8,9 @@ suite('ComputedStylePropertyMap', function() {
...
 
-  test('ComputedStylePropertyMap is a ComputedStylePropertyMap and a ' +
-      'StylePropertyMap', function() {
-    var computedStyleMap = new ComputedStylePropertyMap(this.element);
-    assert.instanceOf(computedStyleMap, ComputedStylePropertyMap);
-    assert.instanceOf(computedStyleMap, typedOMTesting.StylePropertyMap);
-  });
-
-  test('ComputedStylePropertyMap.get works for properties that can only be ' +
+  test('Computed StylePropertyMap.get works for properties that can only be ' +
```
rjwright: Wording here is a little odd. Maybe
"Computed StylePropertyMapReadOnly.get works for properties that can only be NumberValues"
or
"Getting opacity from the computed style map works as expected"

rjwright:
LGTM with comments

src/style-property-map-readonly.js
```diff
@@ -103,27 +92,22 @@
...
       // TODO: Construct the objects for each type of thing and add them to the
       // output array.
     }
     return output;
   };
 
-  function getComputedStyleMap(element) {
-    return new ComputedStylePropertyMap(element);
-  }
+  scope.getComputedStyleMap = function(element) {
```
rjwright: Please make this 

```
getComputedStyleMap = function(element)
```

and put

```
scope.getComputedStyleMap = getComputedStyleMap
```

below (with the other scope member assignments).

src/style-property-map-readonly.js
```diff
@@ -103,27 +92,22 @@
...
+    return new StylePropertyMapReadOnly(getComputedStyle(element));
+  };
 
-  scope.getComputedStyleMap = getComputedStyleMap;
+  internal.StylePropertyMapReadOnly = StylePropertyMapReadOnly;
   if (TYPED_OM_TESTING)
-    testing.ComputedStylePropertyMap = ComputedStylePropertyMap;
+    testing.StylePropertyMapReadOnly = StylePropertyMapReadOnly;
```
rjwright, line 112: One line blocks should have braces.

src/style-property-map.js
```diff
@@ -14,16 +14,25 @@
...
+
+  StylePropertyMap.prototype.append = function(property, value) {
+    throw new TypeError('Function not implemented yet');
+  };
+
+  StylePropertyMap.prototype.delete = function(property) {
+    throw new TypeError('Function not implemented yet');
+  };
+
+  scope.Element.prototype.styleMap = function() {
```
rjwright: Can just be

```
Element.prototype.styleMap
```

Since we haven't explicitly put Element on scope.

test/js/computed-style-property-map.js
```diff
@@ -8,25 +8,17 @@ suite('ComputedStylePropertyMap', function() {
...
+  test('Computed StylePropertyMap.get works for properties that can be NumberValues', function() {
+    var computedStyleMap = getComputedStyleMap(this.element);
     var opacity = computedStyleMap.get('opacity');
     assert.instanceOf(opacity, NumberValue);
     assert.equal(opacity.value, 0.5);
     assert.equal(opacity.cssString, '0.5');
   });
 
-  test('window.getComputedStyleMap works', function() {
+  test('window.getComputedStyleMap method functions correctly', function() {
```
rjwright: This was worded better before.

njlawton:
Yes what you stated above were the purposes of the CL

test/js/inline-style-property-map.js
```diff
@@ -0,0 +1,15 @@
...
+  setup(function() {
+    this.element = document.createElement('div');
+    document.body.appendChild(this.element);
+    this.element.style.opacity = '0.5';
+  });
+  teardown(function() {
+    document.body.removeChild(this.element);
+  });
+
+  test('Test that the Element.styleMap method returns a StylePropertyMap object', function() {
```
rjwright: 'Test that the' is redundant. Remember there are dozens of test cases.

test/js/inline-style-property-map.js
```diff
@@ -0,0 +1,15 @@
...
+    var inlineStyleMap = this.element.styleMap();
+    assert.instanceOf(inlineStyleMap, StylePropertyMap, 'The styleMap method should return an instance of InlineStylePropertyMap');
```
rjwright: s/InlineStylePropertyMap/StylePropertyMap

test/js/computed-style-property-map.js
```diff
@@ -8,16 +8,8 @@ suite('ComputedStylePropertyMap', function() {
...
-      'StylePropertyMap', function() {
-    var computedStyleMap = new ComputedStylePropertyMap(this.element);
-    assert.instanceOf(computedStyleMap, ComputedStylePropertyMap);
-    assert.instanceOf(computedStyleMap, typedOMTesting.StylePropertyMap);
-  });
-
-  test('ComputedStylePropertyMap.get works for properties that can only be ' +
-      'NumberValues', function() {
-    var computedStyleMap = new ComputedStylePropertyMap(this.element);
+  test('Computed StylePropertyMap.get works for properties that can be NumberValues', function() {
```
rjwright: Wording here is a little odd. This is testing a method of StylePropertyMapReadOnly, not StylePropertyMap.

Maybe
"Computed StylePropertyMapReadOnly.get works for properties that can only be NumberValues"
or
"Getting opacity from the computed style map works as expected"

njlawton:
should "Computed StylePropertyMapReadOnly.get works for properties that can only be NumberValues"
just be "Computed StylePropertyMapReadOnly.get works for properties that can be NumberValues"

this was no my test but I think what it is trying to say is that currently NumberValues works and other values will work later

rjwright:
> should "Computed StylePropertyMapReadOnly.get works for properties that can only be NumberValues"
> just be "Computed StylePropertyMapReadOnly.get works for properties that can be NumberValues"

Yep :)

test/js/keyword-value.js
```diff
@@ -33,8 +33,7 @@ suite('KeywordValue', function() {
     assert.isFalse(KeywordValue.isKeywordValue({}));
   });
 
-  test('KeywordValue.cssString and KeywordValue.keywordValue are correctly ' +
-      'initialized', function() {
+  test('KeywordValue.cssString and KeywordValue.keywordValue are correctly initialized', function() {
```
rjwright, line 36: Is this  meant to be in this PR?

PR #35, opened 2015-12-10
Make sure that the faked out inheritance includes the name of the constructor of the superclass.

wilddamon:
- Rename the "baseClasses" namespace to "internal"
- Moved all namespaces except testing into the typedOM namespace
- Add an internal.inherit method which sets up "inheritance" of classes

Now you can see the inheritance chain of class names in the inspector:

```
c = new CalcLength({px:5})
CalcLength {px: 5, percent: null, em: null, ex: null, ch: null…}
ch: null
cm: null
cssString: "calc(5px)"
em: null
ex: null
in: null
mm: null
pc: null
percent: null
pt: null
px: 5
q: null
rem: null
vh: null
vmax: null
vmin: null
vw: null
__proto__: CalcLength
   constructor: CalcLength(dictionary)
   divide: (divider)
   equals: (other)
   multiply: (multiplier)
   __proto__: LengthValue
       _asCalcLength: ()
       add: (addedLength)
       constructor: LengthValue()
       divide: (divider)
       equals: (other)
       multiply: (multiplier)
       parse: (cssString)
       subtract: (subtractedLength)
       __proto__: StyleValue
           constructor: StyleValue()
          __proto__: Object
```

### DIFF
--- a/src/style-value.js
+++ b/src/style-value.js
@@ -31,4 +31,4 @@
   if (TYPED_OM_TESTING)
     testing.StyleValue = StyleValue;
 
-})(baseClasses, window, typedOMTesting);
+})(typedOM.baseClasses, window, typedOMTesting);
--- a/src/style-value.js
+++ b/src/style-value.js
@@ -12,7 +12,7 @@
 //     See the License for the specific language governing permissions and
 // limitations under the License.
 
-(function(shared, scope, testing) {
+(function(internal, scope, testing) {
 
   function StyleValue() {}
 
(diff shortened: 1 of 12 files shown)


src/style-value.js
```diff
@@ -12,24 +12,23 @@
...
-  }
+  function StyleValue() {}
 
   // TODO: Include parsing logic here.
   StyleValue.parse = function(property, value) {
     if (typeof value == 'string') {
       numberValue = Number.parseFloat(value);
       if (numberValue !== NaN) {
-        return new NumberValue(numberValue);
+        return new scope.NumberValue(numberValue);
```
rjwright, line 24: Is this just a stylistic thing? Do we want to be doing this everywhere? I'm wondering if we should even bother having a 'scope' parameter if it's always going to be window.

rjwright:
LGTM

src/style-value.js
wilddamon, line 24: We're doing scope.Thing everywhere else, and I think it should be consistent. But there is certainly room for discussion on the actual utility of having a scope that is window - but lets address it in a new PR.

PR #36, opened 2015-12-10
Remove validation from KeywordValue class

wilddamon:


rjwright:
LGTM & seems sensible. What was the reasoning behind this decision? That checking valid keywords is just too complex?

rjwright:
This will need a rebase though

PR #37, opened 2015-12-14
Small style cleanup, mainly scope stuff

rjwright:
- Functions exposed on scope don't need to be on testing
- Remove unneeded scope arguments
- One-line blocks should have braces around them
- Don't add elements to documentElement for testing

### DIFF
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -12,7 +12,7 @@
 //     See the License for the specific language governing permissions and
 // limitations under the License.
 
-(function(internal, scope, testing) {
+(function(internal, scope) {
 
   // TODO: CalcLength(cssString)
   function CalcLength(dictionary) {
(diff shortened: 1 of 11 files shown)


nattridgell:
LGTM :+1:

For my reference:
- if an object doesn't have a child types (eg. NumberValue, SimpleLength, etc), then it should just be in (internal, scope), not testing, and be declared in scope (ie. scope.Blah = Blah).
- otherwise, if it has child types (eg. StyleValue, LengthValue, TransformComponent), then it should be in testing as well, and declared in internal (ie. internal.Blah = Blah). (Enabling the use of internal.inherit...)

Is this correct?

wilddamon:
LGTM

rjwright:
@nattridgell 
- Put things on `testing` if they aren't already exposed & you need to access them in a test.
- Put things on `scope` if you want to expose them in the API. This makes them accessible everywhere so there is no need also put them on `testing`.
- Put things on `internal` if you want them to be shared between source files but not exposed in the API. This doesn't expose to them tests so they'll need to be put on `testing` to be tested.

PR #38, opened 2015-12-14
Add TransformComponent and Matrix

nattridgell:
Adds:
- TransformComponent
- Very basic TransformComponent tests
- Matrix (for transforms)
- Matrix.multiply functions for multiplying 2D and 3D matrices
  Note: Will be needed by TransformValue
- Matrix.is2DComponent function
- many tests for matrix multiplication using
  precomputed results

### DIFF
--- a/src/matrix.js
+++ b/src/matrix.js
@@ -1,0 +1,123 @@
+// Copyright 2015 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+//     You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(internal, scope, testing) {
+
+  // Note: A Matrix is stored and referenced in column-major order.
+  // TODO: Update the names of parameters (a, b, ... , p) once they are changed
+  // in the specifications and add getters for these parameters.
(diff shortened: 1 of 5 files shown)


src/matrix.js
```diff
@@ -0,0 +1,121 @@
...
+  // Note: A Matrix is stored and referenced in column-major order.
+  // TODO: Update the names of parameters (a, b, ... , p) once they are changed
+  // in the specifications and add getters for these parameters.
+  function Matrix(a, b, c, d, e, f, g, h, i, j, k, l, m, n, o, p) {
+    if (arguments.length != 6 && arguments.length != 16) {
+      throw new TypeError('Expecting 6 or 16 arguments.');
+    }
+
+    this._matrix = [];
+    for (var index = 0; index < arguments.length; index++) {
```
wilddamon, line 26: It's more common just to use "i" as the variable name here.

src/transform-component.js
```diff
@@ -0,0 +1,35 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(internal, testing) {
+
+  function TransformComponent() {
+  }
+
+  TransformComponent.prototype = {
```
wilddamon: I think we were moving to get rid of inline declarations like these... instead please use

TransformComponent.prototype.asMatrix = function() { ... }
TransformComponent.prototype.isDDComponent = function() { ... }

test/js/transform-component.js
```diff
@@ -0,0 +1,13 @@
+suite('TransformComponent', function() {
+  test('TransformComponent is a TransformComponent', function() {
```
wilddamon: This seems redundant... no need to test language features?

src/matrix.js
nattridgell, line 26: Cannot, i, j, and k are defined as parameters.

wilddamon:
LGTM

PR #39, opened 2015-12-14
Created and implemented the PropertyDictionary class

njlawton:
This class contains methods for checking if strings contain vaild CSS propertys as well as methods for testing valid inputs of style values into CSS propertys.
A test class was also created to validate the methods created

### DIFF
--- a/src/property-dictionary.js
+++ b/src/property-dictionary.js
@@ -1,0 +1,95 @@
+// Copyright 2015 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+//     You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(shared, scope, testing) {
+
+  function PropertyDictionary() {
+    this.propertyMap = {
+     'height': [LengthValue],
+     'pitch-range': [NumberValue],
+     'border-top-width': [LengthValue]
+    };
+
+    this.propertyKeywords = {
+      'height': ['auto', 'inherit'],
+      'pitch-range': ['inherit'],
+      'border-top-width': ['inherit']
+    };
+
+    this.allowsPercentage = {
+      'height': true
+    };
+  }
+  var instance;
+
+  PropertyDictionary.prototype.isProperty = function(property) {
+    return (this.propertyMap.hasOwnProperty(property));
+  };
+
+  PropertyDictionary.prototype._lengthValueHasPercentage = function(lengthValue) {
+    if (!(lengthValue instanceof LengthValue)) {
+      return false;
+    }
+
+    if (lengthValue instanceof CalcLength) {
+      return (lengthValue.percent != null);
+    }
+
+    return (lengthValue.type == 'percent');
+  };
+
+  PropertyDictionary.prototype.isValidInput = function(property, styleValue) {
+    if (this.isProperty(property) == false) {
+      return false;
+    }
+
+    if (styleValue instanceof KeywordValue) {
+      if (!(this.propertyKeywords.hasOwnProperty(property))) {
+        return false;
+      }
+      return (this.propertyKeywords[property].indexOf(styleValue.keywordValue) > -1);
+    }
+
+    for (var i = 0; i < this.propertyMap[property].length; i++) {
+      var styleValueType = this.propertyMap[property][i];
+      if (!(styleValue instanceof styleValueType)) {
+        continue;
+      }
+      if (!(styleValue instanceof LengthValue || this._lengthValueHasPercentage(styleValue))) {
+        return true;
+      }
+      if (this.allowsPercentage.hasOwnProperty(property)) {
+        return true;
+      }
+    }
+    return false;
+  };
+
+  var createInstance = function() {
+    var object = new PropertyDictionary();
+    return object;
+  };
+
+  var getInstance = function() {
+    if (!instance) {
+     instance = createInstance();
+    }
+    return instance;
+  };
+
+  shared.getInstance = getInstance;
+  if (TYPED_OM_TESTING)
+    testing.getInstance = getInstance;
+
+})(typedOM.internal, window, typedOMTesting);
--- a/src/property-dictionary.js
+++ b/src/property-dictionary.js
@@ -33,13 +33,13 @@
   }
   var instance;
 
-  PropertyDictionary.prototype.isProperty = function(property) {
+  PropertyDictionary.prototype.isSupportedProperty = function(property) {
     return (this.propertyMap.hasOwnProperty(property));
   };
 
   PropertyDictionary.prototype._lengthValueHasPercentage = function(lengthValue) {
     if (!(lengthValue instanceof LengthValue)) {
-      return false;
+      throw new TypeError('The input to this method must be an object of type LengthValue');
     }
 
     if (lengthValue instanceof CalcLength) {
@@ -50,7 +50,7 @@
   };
 
   PropertyDictionary.prototype.isValidInput = function(property, styleValue) {
-    if (this.isProperty(property) == false) {
+    if (this.isSupportedProperty(property) == false) {
       return false;
     }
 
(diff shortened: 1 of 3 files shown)


src/property-dictionary.js
```diff
@@ -0,0 +1,95 @@
...
+      'border-top-width': ['inherit']
+    };
+
+    this.allowsPercentage = {
+      'height': true
+    };
+  }
+  var instance;
+
+  PropertyDictionary.prototype.isProperty = function(property) {
```
wilddamon: What is the purpose of this function? Is it to check that the property is a supported property? If so, please change the name to "isSupportedProperty".

src/property-dictionary.js
```diff
@@ -0,0 +1,95 @@
...
+    return (this.propertyMap.hasOwnProperty(property));
+  };
+
+  PropertyDictionary.prototype._lengthValueHasPercentage = function(lengthValue) {
+    if (!(lengthValue instanceof LengthValue)) {
+      return false;
```
wilddamon: Perhaps this should throw an exception to make it clear that the caller is doing something silly.

src/property-dictionary.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    if (lengthValue instanceof CalcLength) {
+      return (lengthValue.percent != null);
+    }
+
+    return (lengthValue.type == 'percent');
```
wilddamon: Can you use the enum value instead of writing the string directly here?

src/property-dictionary.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  PropertyDictionary.prototype.isValidInput = function(property, styleValue) {
+    if (this.isProperty(property) == false) {
```
wilddamon: if (!this.isProperty(property)) {

src/property-dictionary.js
```diff
@@ -0,0 +1,95 @@
...
+      return false;
+    }
+
+    if (styleValue instanceof KeywordValue) {
+      if (!(this.propertyKeywords.hasOwnProperty(property))) {
+        return false;
+      }
+      return (this.propertyKeywords[property].indexOf(styleValue.keywordValue) > -1);
```
wilddamon: how about making a helper method for this check

```
PropertyDictionary.prototype.isValidKeyword = function(property, str) {
  return this.propertyKeywords[property].indexOf(str) >= 0;
}
```

then this function becomes

```
if (styleValue instanceof KeywordValue) {
  return this.isProperty(property) ? this.isValidKeyword(styleValue.keywordValue) : false;
}
```

src/property-dictionary.js
```diff
@@ -0,0 +1,95 @@
...
+  };
+
+  var createInstance = function() {
+    var object = new PropertyDictionary();
+    return object;
+  };
+
+  var getInstance = function() {
+    if (!instance) {
+     instance = createInstance();
```
wilddamon: The constructor for PropertyDictionary is quite simple, so it should be the same to delete createInstance, and make this function read

```
var getInstance = function() {
  if (!instance) {
    instance = new PropertyDictionary();
  }
  return instance;
}
```

src/property-dictionary.js
```diff
@@ -0,0 +1,95 @@
...
+    }
+    return instance;
+  };
+
+  shared.getInstance = getInstance;
```
wilddamon: This function name is too general - what if we want some other singleton class?

How about something like "getPropertyDictionaryInstance()", or as is common in blink, "propertyDictionary()"

test/js/property-dictionary.js
```diff
@@ -0,0 +1,54 @@
...
+  test('The _lengthValueHasPercentage method should return true if a SimpleLength object has a type of percent', function() {
+    assert.isTrue(propertyDictionary._lengthValueHasPercentage(new SimpleLength(10, 'percent')));
+  });
+
+  test('The isValidInput method should return true if the property can accept the style value entered', function() {
+    assert.isTrue(propertyDictionary.isValidInput('height', new SimpleLength(9.2, 'px')));
+  });
+
+  test('Testing the isValidInput method with a non Length value property' , function() {
+    assert.isTrue(propertyDictionary.isValidInput('pitch-range', new NumberValue(5)));
```
wilddamon: Just add this line to the above test - it's essentially testing the same thing as

`The isValidInput method should return true if the property can accept the style value entered`

test/js/property-dictionary.js
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  test('The isValidInput method should return false if the property cant accept the style value entered' , function() {
```
wilddamon: s/cant/can't

src/property-dictionary.js
```diff
@@ -0,0 +1,94 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(shared, scope, testing) {
+
+  function PropertyDictionary() {
+    this.propertyMap = {
```
wilddamon: Can we change the name of this to be "validProperties" for clarity? It's not very helpful to have a PropertyDictionary that contains a propertyMap...

src/property-dictionary.js
```diff
@@ -0,0 +1,94 @@
...
+      }
+      return false;
+    }
+
+    for (var i = 0; i < this.propertyMap[property].length; i++) {
+      var styleValueType = this.propertyMap[property][i];
+      if (!(styleValue instanceof styleValueType)) {
+        continue;
+      }
+      if (!(styleValue instanceof LengthValue)) {
```
wilddamon: The logic here is confusing. I know you're trying to reduce nesting, but in this case all the negation makes it more confusing... how about something like

```
...
  for (var i = 0; i < this.propertyMap[property].length; i++) {
    var styleValueType = this.propertyMap[property][i];
    if (styleValue instanceOf styleValueType) {
      if (styleValueType != LengthValue) {
        return true;
      }
      return (!this._lengthValueHasPercentage(styleValue) || this.allowsPercentage[property]);
    }
  }
  return false;
};
```

FWIW, you can also return false if you find a percentage LengthValue when allowsPercentage returns false.

Also you don't need to check hasOwnProperty for allowsPercentage, you can just do allowsPercentage[property], which returns `undefined`, which is "falsy" (equivalent to `false`).

src/property-dictionary.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+  var instance;
+
+  PropertyDictionary.prototype.isSupportedProperty = function(property) {
+    return (this.propertyMap.hasOwnProperty(property));
+  };
+
+  PropertyDictionary.prototype._lengthValueHasPercentage = function(lengthValue) {
+    if (!(lengthValue instanceof LengthValue)) {
+      throw new TypeError('The input to this method must be an object of type LengthValue');
```
wilddamon: It's better to say

'The input to _lengthValueHasPercentage must be an object of type LengthValue'

for clarity, since you know exactly where to go looking for you error in that case.

src/property-dictionary.js
wilddamon: These three maps should be "private", you can indicate this by prepending their names with an underscore;

this._validProperties
this._validKeywords
this._allowsPercentage

njlawton:
Finished those changes are there any other major problems? otherwise I will work on the next pull request and merge before submitting pull request for inspection.

src/property-dictionary.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+  var instance;
+
+  PropertyDictionary.prototype.isSupportedProperty = function(property) {
+    return (this._validProperties.hasOwnProperty(property));
+  };
+
+  PropertyDictionary.prototype._lengthValueHasPercentage = function(lengthValue) {
+    if (!(lengthValue instanceof LengthValue)) {
+      throw new TypeError('The input to _lengthValueHasPercentage must be an object of type LengthValue');
```
wilddamon: Line too long, break after (

src/property-dictionary.js
```diff
@@ -0,0 +1,91 @@
...
+      return false;
+    }
+
+    for (var i = 0; i < this._validProperties[property].length; i++) {
+      var styleValueType = this._validProperties[property][i];
+      if (styleValue instanceof styleValueType) {
+        if (styleValueType != LengthValue) {
+          return true;
+        }
+        return (!this._lengthValueHasPercentage(styleValue) || this._allowsPercentage.hasOwnProperty(property));
```
wilddamon: Line too long, break after ||

src/property-dictionary.js
```diff
@@ -0,0 +1,91 @@
...
+
+  var propertyDictionary = function() {
+    if (!instance) {
+      instance = new PropertyDictionary();
+    }
+    return instance;
+  };
+
+  shared.propertyDictionary = propertyDictionary;
+  if (TYPED_OM_TESTING)
```
wilddamon: needs braces {

wilddamon:
Three really minor comments, otherwise LGTM

njlawton:
Some lines i.e. defining methods are still slightly too long, also the line with the error message I broke is still slightly too long (none more than 90 characters) is this ok?

wilddamon:
LGTM

PR #40, opened 2015-12-15
Add Scale (a TransformComponent)

nattridgell:
Key points for Scale:
- matrix constructed for scale
  - 2D: diagonal matrix with (x,y) on diagonal
  - 3D: diagonal matrix with (x,y,z,1) on diagonal
- z attribute is null if it is 2d component
- uses scale / scale3d cssString (as appropriate)

### DIFF
--- a/src/scale.js
+++ b/src/scale.js
@@ -1,0 +1,61 @@
+// Copyright 2015 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


wilddamon:
LGTM

I think what you have here looks fine, where z is null for a 2D instance. The logic gets a little more complicated if you make it a number, and the caller can promote it to a 3D matrix through the matrix class if they really want to.

PR #41, opened 2015-12-15
Add Skew (a TransformComponent)

nattridgell:
Key points for Skew:
- always 2d component
- constructor takes (ax, ay), which are both angles in degrees.
- 2D matrix constructed in column-major order is:
  [1, tan(ay), tan(ax), 1, 0, 0]

Note: for all TransformComponents, use internal methods in constructor:
- _computeMatrix
- _generateCssString


wilddamon:
LGTM

PR #42, opened 2015-12-16
Make LengthValue.LengthType an enum again. Add helper functions.

wilddamon:
- Makes LengthValue.LengthType more like an enum using a JS object.
- Adds new helper methods for manipulating objects:

```
internal.objects.foreach(object, callback, opt_thisObject)
internal.objects.any(object, callback, opt_thisObject)
```


src/util.js
```diff
@@ -12,13 +12,58 @@
...
 
-(function(internal) {
+(function(internal, testing) {
 
   function inherit(childCtor, parentCtor) {
     childCtor.prototype = Object.create(parentCtor.prototype);
     childCtor.prototype.constructor = childCtor;
   }
 
+  /*
```
rjwright, line 22: Should stick to line comment delimiters: `// comment`.

src/util.js
```diff
@@ -12,13 +12,58 @@
...
+   * Calls the callback for each value in the object.
+   * @param {!Object} object: The object to iterate over.
+   * @param {Function} callback: A function to be called for each value in the
+   *   object. It should be specified as function(value, key, inputObject).
+   *   Its return value is ignored.
+   * @param {Object} opt_this: The value that will appear as 'this' within the
+   *   callback.
+   */
+  function foreach(object, callback, opt_this) {
```
rjwright, line 31: Underscore in var name? I thought we were avoiding that.

src/scope.js
```diff
@@ -13,7 +13,11 @@
 // limitations under the License.
 
 typedOM = {};
+/* Non-exposed subclasses and helper methods. */
```
rjwright: Switch to `//` delimiters

rjwright:
LGTM with nits

src/util.js
wilddamon, line 31: This is a convention in Google JavaScript, optional params are denoted as with opt + underscore + camelcase like this: opt_camelCase. Do we want to keep this or make it optCamelCase for optional params?

src/util.js
wilddamon, line 22: This is proper JSDoc, which I thought would be appropriate for this location. In Google we use the /*\* style comment for function documentation - do we want to stick with that or change to the other kind?

rjwright, line 22: I prefer it to be consistent if we're not actually planning to do JSDoc but it's not a big deal. My nits are mainly based on things Doug nitpicked about when I started on Web Animations :P

src/util.js
rjwright, line 31: Na, leave it like this. That's a good conventions, I've just never seen it before.

PR #43, opened 2015-12-16
StylePropertyMap set method implemented 

njlawton:
The set method has been implemented for longhand CSS properties that are defined by only one StyleValue

### DIFF
--- a/src/style-property-map.js
+++ b/src/style-property-map.js
@@ -27,16 +27,19 @@
 
     if (value instanceof StyleValue) {
       if (!cssPropertyDictionary.isValidInput(property, value)) {
-        throw new TypeError('This StyleValue type cannot be set to this property');
+        throw new TypeError(
+          'This StyleValue type cannot be set to this property');
       }
       this._styleObject[property] = value.cssString;
       return;
     }
 
     if (value instanceof Array) {
-      throw new TypeError('Setting a sequence of StyleValues is not implemented yet');
+      throw new TypeError(
+        'Setting a sequence of StyleValues is not implemented yet');
     }
-    throw new TypeError('The value you are setting must be a StyleValue object');
+    throw new TypeError(
+      'The value you are setting must be a StyleValue object');
   };
 
   StylePropertyMap.prototype.append = function(property, value) {
--- a/src/style-property-map.js
+++ b/src/style-property-map.js
@@ -34,12 +34,14 @@
       throw new TypeError(
         'The value must be a StyleValue or sequence of StyleValues');
     }
-      
+
     if (!cssPropertyDictionary.isValidInput(property, value)) {
-      if(value instanceof KeywordValue) {
-        throw new TypeError(property + ' does not take the keyword ' + value.cssString);
+      if (value instanceof KeywordValue) {
+        throw new TypeError(property +
+          ' does not take the keyword ' + value.cssString);
       }
-      throw new TypeError(property + ' does not take values of type ' + value.constructor.name);
+      throw new TypeError(property +
+        ' does not take values of type ' + value.constructor.name);
     }
     this._styleObject[property] = value.cssString;
   };
--- a/src/style-property-map.js
+++ b/src/style-property-map.js
@@ -20,7 +20,23 @@
   internal.inherit(StylePropertyMap, internal.StylePropertyMapReadOnly);
 
   StylePropertyMap.prototype.set = function(property, value) {
-    throw new TypeError('Function not implemented yet');
+    var cssPropertyDictionary = propertyDictionary();
+    if (!cssPropertyDictionary.isSupportedProperty(property)) {
+      throw new TypeError(property + 'Is not a Supported CSS property');
+    }
+
+    if (value instanceof StyleValue) {
+      if (!cssPropertyDictionary.isValidInput(property, value)) {
+        throw new TypeError('This StyleValue type cannot be set to this property');
+      }
+      this._styleObject[property] = value.cssString;
+      return;
+    }
+
+    if (value instanceof Array) {
+      throw new TypeError('Setting a sequence of StyleValues is not implemented yet');
+    }
+    throw new TypeError('The value you are setting must be a StyleValue object');
   };
 
   StylePropertyMap.prototype.append = function(property, value) {
--- a/src/style-property-map.js
+++ b/src/style-property-map.js
@@ -22,24 +22,26 @@
   StylePropertyMap.prototype.set = function(property, value) {
     var cssPropertyDictionary = propertyDictionary();
     if (!cssPropertyDictionary.isSupportedProperty(property)) {
-      throw new TypeError(property + 'Is not a Supported CSS property');
-    }
-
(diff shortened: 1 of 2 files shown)


src/style-property-map.js
```diff
@@ -20,7 +20,26 @@
   internal.inherit(StylePropertyMap, internal.StylePropertyMapReadOnly);
 
   StylePropertyMap.prototype.set = function(property, value) {
-    throw new TypeError('Function not implemented yet');
+    var cssPropertyDictionary = propertyDictionary();
+    if (!cssPropertyDictionary.isSupportedProperty(property)) {
+      throw new TypeError(property + 'Is not a Supported CSS property');
```
wilddamon: Nitpick: Need a space and should be a lowercase "i" for "is".

``` javascript
throw new TypeError(property + ' is not a supported CSS property.');
```

src/style-property-map.js
```diff
@@ -20,7 +20,26 @@
...
+    }
+
+    if (value instanceof StyleValue) {
+      if (!cssPropertyDictionary.isValidInput(property, value)) {
+        throw new TypeError(
+          'This StyleValue type cannot be set to this property');
```
wilddamon: It's better to include the property and value in the error message.

``` javascript
throw new TypeError(property + ' does not take values of type ' + value.constructor.name);
```

src/style-property-map.js
```diff
@@ -20,7 +20,26 @@
...
+      this._styleObject[property] = value.cssString;
+      return;
+    }
+
+    if (value instanceof Array) {
+      throw new TypeError(
+        'Setting a sequence of StyleValues is not implemented yet');
+    }
+    throw new TypeError(
+      'The value you are setting must be a StyleValue object');
```
wilddamon: "... must be a StyleValue or sequence of StyleValues"?

test/js/inline-style-property-map.js
```diff
@@ -8,8 +8,44 @@ suite('Inline StylePropertyMap', function() {
     document.body.removeChild(this.element);
   });
 
-  test('Element.styleMap method returns a StylePropertyMap object', function() {
+  test('the Element.styleMap method returns a StylePropertyMap object for that element', function() {
```
wilddamon: Nitpick: Should start with a capital letter.

src/style-property-map.js
```diff
@@ -20,7 +20,26 @@
...
+    if (value instanceof StyleValue) {
+      if (!cssPropertyDictionary.isValidInput(property, value)) {
+        throw new TypeError(
+          'This StyleValue type cannot be set to this property');
+      }
+      this._styleObject[property] = value.cssString;
+      return;
+    }
+
+    if (value instanceof Array) {
```
wilddamon, line 28: How about we put this first, then it can eventually be passed off to a helper function. Then you can get rid of the floating `return` statement and write something like

``` javascript
StylePropertyMap.prototype.set = function(property, value) {
  var cssPropertyDictionary = propertyDictionary();
  if (!cssPropertyDictionary.isSupportedProperty(property)) {
    throw new TypeError(property + ' is not a supported CSS property.');
  }

  if (value instanceof Array) {
    // pass to helper function
  }

  if (!value instanceof StyleValue) {
    throw new TypeError('value must be a StyleValue or sequence of StyleValues');
  }
  if (!cssPropertyDictionary.isValidInput(property, value)) {
    throw new TypeError(property + ' does not take values of type ' + value.constructor.name);
  }

  this._styleObject[property] = value.cssString;
}
```

test/js/inline-style-property-map.js
```diff
@@ -8,8 +8,44 @@ suite('Inline StylePropertyMap', function() {
...
     var inlineStyleMap = this.element.styleMap();
-    assert.instanceOf(inlineStyleMap, StylePropertyMap, 'The styleMap method should return an instance of StylePropertyMap');
+    assert.instanceOf(inlineStyleMap, StylePropertyMap);
+  });
+
+  test('The set method should correctly set a supported StyleValue to a CSS style property of an HTML element', function() {
```
wilddamon: This title is a bit confusing, how about

``` javascript
test('The set method successfully sets the CSS string of the StyleValue on an element.', function() {
```

test/js/inline-style-property-map.js
```diff
@@ -8,8 +8,44 @@ suite('Inline StylePropertyMap', function() {
...
+
+    assert.strictEqual(this.element.style['height'], simpleLength.cssString);
+  });
+
+  test('The set method should correctly set a supported KeywordValue to a CSS style property of an HTML element', function() {
+    var inlineStyleMap = this.element.styleMap();
+    var keyword = new KeywordValue('auto');
+    inlineStyleMap.set('height', keyword);
+
+    assert.strictEqual(this.element.style['height'], keyword.cssString);
```
wilddamon: This test is actually testing the same thing as the above test. You can delete it.

test/js/inline-style-property-map.js
```diff
@@ -8,8 +8,44 @@ suite('Inline StylePropertyMap', function() {
...
+  });
+
+  test('The set method should throw a TypeError if a StyleValue unsupported by the CSS style property is set', function() {
+    var inlineStyleMap = this.element.styleMap();
+    var numberValue = new NumberValue(42);
+
+    assert.throw(function() {inlineStyleMap.set('height', numberValue)}, TypeError);
+  });
+
+  test('The set method should throw a TypeError if a KeywordValue unsupported by the CSS style property is set ', function() {
```
wilddamon, line 31: A KeywordValue is a StyleValue, so this test is testing the same thing as the one above it. It can also be deleted.

test/js/inline-style-property-map.js
```diff
@@ -8,8 +8,44 @@ suite('Inline StylePropertyMap', function() {
...
+
+    assert.throw(function() {inlineStyleMap.set('height', keyword)}, TypeError);
+  });
+
+  test('The set method should throw a TypeError if a non StyleValue is inputed into the function', function() {
+    var inlineStyleMap = this.element.styleMap();
+
+    assert.throw(function() {inlineStyleMap.set('height', 4)}, TypeError);
   });
 });
```
wilddamon, line 49: Please also add a test for setting an unsupported property, something like

``` javascript
inlineStyleMap.set('lemons', new SimpleLength(3, 'px'));
```

njlawton:
I have changed how error messages are displayed like you said but due to getting confusing messages like height does not take values of type KeywordValue I have separated out the errors into two so that if an incorrect keyword is entered it will say something like property + ' does not take the keyword ' + value.cssString as a result I have not removed (and slightly updated) the test case you asked me to remove

wilddamon:
Oh I see. Perhaps a note in the test name that it is for StyleValues that are not KeywordValues then would clear up any confusion.

njlawton:
As I changed the wording of the test case above it to fix the problem the box is not going away

wilddamon:
LGTM

The comment still shows up because you didn't change that line. Which is as expected so no worries :)

PR #44, opened 2015-12-16
Add TransformValue

nattridgell:
Key points for TransformValue:
- constructor
  - takes an array of TransformComponent values
  - must have at least 1 value
- is2D if it's matrix is a 2D component,
  that is, if all the values are 2D
- matrix is formed by cross multiplying all components
  transform matrices in order
- cssString is a space seperated list of cssStrings from
  each transform component in order

### DIFF
--- a/test/js/transform-value.js
+++ b/test/js/transform-value.js
@@ -35,7 +35,7 @@ suite('TransformValue', function() {
     assert.doesNotThrow(function() {transform = new TransformValue(values)});
     assert.isTrue(transform.is2D());
     assert.strictEqual(transform.cssString,
-        values[0].cssString + ' ' + values[1].cssString);
+        scale.cssString + ' ' + scale.cssString);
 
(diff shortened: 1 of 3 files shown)


test/js/transform-value.js
```diff
@@ -0,0 +1,103 @@
...
+
+  test('TransformValue constructor works with duplicate component types',
+        function() {
+    var transform;
+    var scale = new Scale(2, -1);
+    var values = [scale, scale];
+    assert.doesNotThrow(function() {transform = new TransformValue(values)});
+    assert.isTrue(transform.is2D());
+    assert.strictEqual(transform.cssString,
+        values[0].cssString + ' ' + values[1].cssString);
```
wilddamon: you can write 

``` javascript
assert.strictEqual(transform.cssString, scale.cssString + ' ' + scale.cssString)
```

To emphasize that the component is duplicate.

wilddamon:
LGTM

PR #46, opened 2015-12-16
Add Translation (a TransformComponent)

nattridgell:
Key points for a translation:
- constructor
  - accepts 2 or 3 LengthValues as arguments
  - currently only SimpleLengths with type `px` are supported
    (This is due to other types needing to be resolved.)
- z is null if it is a 2D component
- matrix is an identity matrix except for the last column
  - has [x, y] as the last column for 2D
  - has [x, y, z, 1] as the last column for 3D
  - where x, y, z are the px lengths of the length values
- cssString
  - either translate / translate3d
  - has a comma seperated list of the length value cssStrings


src/translation.js
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    for (var i = 0; i < arguments.length; i++) {
+      if (!(arguments[i] instanceof LengthValue)) {
+        throw new TypeError('Translation arguments must be instances of ' +
+            'LengthValue.');
+      } else if (!(arguments[i] instanceof SimpleLength) ||
+          arguments[i].type != 'px') {
+        throw new TypeError('Unsupported LengthValue for Translation. ' +
+            'Only SimpleLength instances with type \'px\' are supported.');
```
wilddamon: Can you just skip the first check then if they have to be SimpleLengths anyway?

test/js/translation.js
```diff
@@ -0,0 +1,80 @@
...
+      function() {
+    var validLength = new SimpleLength(3, 'px');
+    var emSimpleLength = new SimpleLength(3, 'em');
+    var pxCalcLength = new CalcLength({px: 3});
+    assert.throws(function() {new Translation(validLength, emSimpleLength)});
+    assert.throws(function() {new Translation(emSimpleLength, emSimpleLength)});
+    assert.throws(function() {new Translation(validLength, pxCalcLength)});
+    assert.throws(function() {new Translation(pxCalcLength, pxCalcLength)});
+    assert.throws(function() {new Translation(validLength)});
+    assert.throws(function() {new Translation(validLength, validLength, null)});
```
wilddamon, line 35: maybe add some comments for why these are invalid? (the last two are a little confusing since they only contain the variable "validLength")

src/translation.js
```diff
@@ -0,0 +1,72 @@
...
+    this.cssString = this._generateCssString();
+  }
+  internal.inherit(Translation, internal.TransformComponent);
+
+  Translation.prototype.asMatrix = function() {
+    return this._matrix;
+  };
+
+  Translation.prototype._computeMatrix = function() {
+    // Translation represented by the identity matrix with the translation
```
rjwright: I'd change this to 'Translation is represented', just for clarity

test/js/translation.js
```diff
@@ -0,0 +1,82 @@
+suite('Translation', function() {
+  test('Translation is a Translation and TransformComponent', function() {
+    var validLength = new SimpleLength(1, 'px');
+    var translation = new Translation(validLength, validLength);
+    assert.instanceOf(translation, Translation,
```
rjwright: No need for this. You just constructed it so you know this will pass :)

test/js/translation.js
```diff
@@ -0,0 +1,80 @@
...
+    assert.instanceOf(translation, TransformComponent,
+        'A new Translation should be an instance of TransformComponent');
+  });
+
+  test('Translation constructor throws exception for invalid types',
+      function() {
+    assert.throws(function() {new Translation()});
+    assert.throws(function() {new Translation({})});
+    assert.throws(function() {new Translation('1px', '2px')});
+    assert.throws(function() {new Translation(null, null)});
```
wilddamon, line 16: What about a check for an invalid third argument, here or another test method?

wilddamon:
LGTM other than comment about tests

wilddamon:
LGTM

PR #47, opened 2015-12-16
cssString and _matrix constructor helper functions

nattridgell:
- cssString and _matrix are assigned in the constructor
- cssString is generated using _generateCssString function
- _matrix is computed using _computeMatrix function
  (for transforms only)
- NOTE:
  - CalcLength is not included here, as updated in different
    commit
  - Small cssStrings, such as KeywordValue, and NumberValue,
    are not included.

### DIFF
--- a/src/matrix.js
+++ b/src/matrix.js
@@ -31,7 +31,7 @@
       this._matrix[index] = arguments[index];
     }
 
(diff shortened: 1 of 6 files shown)


wilddamon:
LGTM

PR #48, opened 2015-12-16
StylePropertyMap implemented the delete and has methods

njlawton:
Implemented methods and added test cases for them

### DIFF
--- a/src/style-property-map.js
+++ b/src/style-property-map.js
@@ -22,7 +22,7 @@
   StylePropertyMap.prototype.set = function(property, value) {
     var cssPropertyDictionary = propertyDictionary();
     if (!cssPropertyDictionary.isSupportedProperty(property)) {
-      throw new TypeError(property + ' is not a supported CSS property');
+      throw new TypeError('Cannot set ' + property + ' because it is not a supported CSS property');
     }
 
     if (value instanceof Array) {
@@ -53,7 +53,7 @@
   StylePropertyMap.prototype.delete = function(property) {
     var cssPropertyDictionary = propertyDictionary();
     if (!cssPropertyDictionary.isSupportedProperty(property)) {
-      throw new TypeError(property + ' is not a supported CSS property');
+      throw new TypeError('Cannot delete ' + property + ' because it is not a supported CSS property');
     }
     this._styleObject[property] = '';
   };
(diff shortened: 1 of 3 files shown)


src/style-property-map.js
```diff
@@ -22,7 +22,7 @@
   StylePropertyMap.prototype.set = function(property, value) {
     var cssPropertyDictionary = propertyDictionary();
     if (!cssPropertyDictionary.isSupportedProperty(property)) {
-      throw new TypeError(property + 'is not a supported CSS property');
+      throw new TypeError(property + ' is not a supported CSS property');
```
wilddamon: It's generally good practice to have different error messages if they appear more than once - that way you know exactly where to look if you see that message. So in this case you can say

"Cannot set " + property + " because it is not a supported CSS property"
"Cannot delete " + property + ... etc

test/js/inline-style-property-map.js
```diff
@@ -46,4 +46,32 @@ suite('Inline StylePropertyMap', function() {
 
     assert.throw(function() {inlineStyleMap.set('lemons', new SimpleLength(3, 'px'))}, TypeError);
   });
+
+  test('The delete method removes any StyleValue assigned to the CSS style property inputed', function() {
```
wilddamon: The delete method clears the given property.

"inputed" is not a word.

test/js/inline-style-property-map.js
```diff
@@ -46,4 +46,32 @@ suite('Inline StylePropertyMap', function() {
...
+    var inlineStyleMap = this.element.styleMap();
+    this.element.style.height = '10px';
+    inlineStyleMap.delete('height');
+
+    assert.strictEqual(this.element.style['height'], '');
+  });
+
+  test('The delete method should throw a TypeError if an unsupported property is given as input', function() {
```
wilddamon: For brevity, you can just write 

'The delete method should throw a TypeError if an unsupported property'

src/style-property-map.js
```diff
@@ -51,7 +51,19 @@
   };
 
   StylePropertyMap.prototype.delete = function(property) {
-    throw new TypeError('Function not implemented yet');
+    var cssPropertyDictionary = propertyDictionary();
+    if (!cssPropertyDictionary.isSupportedProperty(property)) {
```
wilddamon: As in other PR, can be reduced to 

``` javascript
```

wilddamon:
Minor nitpick, otherwise LGTM

wilddamon:
LGTM

PR #49, opened 2015-12-16
Add Perspective (a TransformComponent)

nattridgell:
Key points for Perspective
- constructor
  - takes a single SimpleLength with type `px`
  - the length must be strictly positive ( > 0 )
- matrix
  - 3D identity + ( -1 / length ) at 4th row, 3rd column
- is2DComponent
  - always false (perspective is a 3D component)
- cssString is for example `perspective(10px)`


src/perspective.js
```diff
@@ -0,0 +1,52 @@
...
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(internal, scope) {
+
+  function Perspective(length) {
+    if (arguments.length != 1) {
+      throw new TypeError('Perspective takes exactly 1 argument.');
+      // TODO: Change check of 'px' when LengthValue.LengthType enum is updated.
+    } else if (!(length instanceof SimpleLength) || length.type != 'px') {
```
wilddamon: Since the later ifs can't be reached if one of the others is matched, these can be

``` javascript
if (condition) {
  throw ...
}
if (condition) {
  throw
}
```

This is enforced by the style guide in C++ so may as well get used to it now :)

nattridgell: Thanks for the note :-) Will do.

src/perspective.js
```diff
@@ -0,0 +1,55 @@
...
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(internal, scope) {
+
+  function Perspective(length) {
+    if (arguments.length != 1) {
+      throw new TypeError('Perspective takes exactly 1 argument.');
+    }
+    // TODO: Change check of 'px' when LengthValue.LengthType enum is updated.
```
wilddamon: This has been landed now.

wilddamon:
LGTM

PR #50, opened 2015-12-17
Add Rotation (a TransformComponent)

nattridgell:
Key points for rotation
- constructor
  - takes (angle) or (angle, x, y, z)
  - arguments are numbers
  - (x, y, z) are all null if 2D component
- matrix
  - relatively complex, view documentation
  - must normalise (x, y, z) first before useing the 3d matrix
    construction in the documentation
- cssString
  - must have text `deg` next to the angle
  - rotate3d order of arguments is (x, y, z, angle)


wilddamon:
LGTM

PR #51, opened 2015-12-20
StylePropertyMap implemented the append method

njlawton:


src/property-dictionary.js
```diff
@@ -18,24 +18,43 @@
...
+      'opacity': ['initial', 'inherit'],
+      'animationIterationCount': ['infinite']
     };
 
     this._allowsPercentage = {
       'height': true
     };
+
+    this._allowsListValues = {
+      'animationIterationCount': ', '
```
wilddamon: What does the comma mean here? Should it just be true like _allowsPercentage?

[edit] Oh I see. To prevent conclusion, perhaps this should be called _listSeparator or something similar.

src/property-dictionary.js
```diff
@@ -18,24 +18,43 @@
...
 
+  PropertyDictionary.prototype.isListValuedProperty = function(property) {
+    return (this._allowsListValues.hasOwnProperty(property));
+  };
+
+  PropertyDictionary.prototype.getListValueSeparator = function(property) {
+    if (this.isListValuedProperty(property)) {
+      return this._allowsListValues[property];
+    }
+    throw new TypeError(property + ' does not support lists of styleValues');
```
wilddamon: s/styleValues/StyleValues (capitalization)

src/style-property-map.js
```diff
@@ -36,22 +36,58 @@
...
+
+    var cssAppendString = this._styleObject[property];
+    var valueSeparator = cssPropertyDictionary.getListValueSeparator(property);
+    for (var i = 0; i < values.length; i++) {
+      if (!cssPropertyDictionary.isValidInput(property, values[i])) {
+        cssPropertyDictionary.throwInvalidInputError(property, values[i]);
+      }
+      if (cssAppendString == '') {
+        cssAppendString += values[i].cssString;
+        continue;
```
wilddamon: Perhaps this should be if/else to prevent confusion. The `continue` statement threw me a little.

src/style-property-map.js
```diff
@@ -36,22 +36,58 @@
...
+      }
+      cssAppendString += valueSeparator + values[i].cssString;
+    }
+    return this._styleObject[property] = cssAppendString;
   };
 
   StylePropertyMap.prototype.delete = function(property) {
-    throw new TypeError('Function not implemented yet');
+    var cssPropertyDictionary = propertyDictionary();
+    if (!cssPropertyDictionary.isSupportedProperty(property)) {
```
wilddamon: You can simplify these two lines by writing

``` javascript
```

It's basically the same length...

src/style-property-map.js
```diff
@@ -36,22 +36,58 @@
...
+      throw new TypeError(property + ' is not a supported CSS property');
+    }
+    this._styleObject[property] = '';
+  };
+
+  StylePropertyMap.prototype.has = function(property) {
+    var cssPropertyDictionary = propertyDictionary();
+    if (!cssPropertyDictionary.isSupportedProperty(property)) {
```
wilddamon: Same here.

src/style-property-map.js
```diff
@@ -36,22 +36,58 @@
...
+      throw new TypeError(property + ' is not a supported CSS property');
+    }
+    return !(this._styleObject[property] == '');
```
wilddamon: You can just write

``` javascript
return !!this._styleObject[property]
```

"", null, undefined and 0 all evaluate to false, most other things evaluate to true, and using `!!` coerces the value to be a boolean.

src/property-dictionary.js
```diff
@@ -19,25 +19,42 @@
      'height': [LengthValue],
      'pitch-range': [NumberValue],
      'border-top-width': [LengthValue],
-     'opacity': [NumberValue]
+     'opacity': [NumberValue],
+     'animationIterationCount': [NumberValue]
```
wilddamon: to be consistent, use `animation-iteration-count`

wilddamon:
LGTM

PR #54, opened 2015-12-21
Copy constructors

nattridgell:
Make copy constructors for:
- SimpleLength (in the specs)
- CalcLength (already implemented, just added tests)
- LengthValue (not in the specs, but needed in PositionValue)

Use the copy constructors in objects storing LengthValues,
so that they store a copy instead.
- SimpleLength copy constructor used in: Perspective, Translation
- LengthValue copy constructor used in: PositionValue

### DIFF
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -14,7 +14,19 @@
 
 (function(internal, testing) {
 
-  function LengthValue() {}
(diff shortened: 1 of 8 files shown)


wilddamon:
LGTM

PR #55, opened 2015-12-21
StylePropertyMap: Implemented setting sequences of StyleValues in the set method

njlawton:


test/js/inline-style-property-map.js
```diff
@@ -21,6 +21,16 @@ suite('Inline StylePropertyMap', function() {
...
   });
 
+  test('The set method successfully sets the CSS string of the list of StyleValues on property ' +
+    'if it supports setting lists of values', function() {
+    var inlineStyleMap = this.element.styleMap();
+    var valueArray = [new NumberValue(4), new NumberValue(5), new KeywordValue('infinite')];
+    this.element.style['animation-iteration-count'] = 'infinite, 2, 5';
+    inlineStyleMap.set('animation-iteration-count', valueArray);
+
+    assert.strictEqual(this.element.style['animation-iteration-count'], '4, 5, infinite');
```
wilddamon, line 30: Should also test that attempting to set a list of values doesn't work for properties that don't support sequences.

test/js/inline-style-property-map.js
```diff
@@ -21,6 +21,16 @@ suite('Inline StylePropertyMap', function() {
     assert.strictEqual(this.element.style['height'], simpleLength.cssString);
   });
 
+  test('The set method successfully sets the CSS string of the list of StyleValues on property ' +
+    'if it supports setting lists of values', function() {
```
wilddamon: For brevity:
"Set successfully sets the style string for the list of StyleValues on a property that supports sequences."

wilddamon:
LGTM

PR #56, opened 2015-12-21
Fixed looking up animation-iteration-count in PropertyDictionary by test cases

njlawton:
animation-iteration-count string was changed in PropertyDictionary but not its test cases this has now been changed and test will pass


wilddamon:
LGTM

PR #57, opened 2015-12-22
Add parsing for LengthValue

wilddamon:
Tests now included.

### DIFF
--- a/src/parsing.js
+++ b/src/parsing.js
@@ -31,7 +31,7 @@
     if (!/^[^(]*$|^calc/.test(string)) {
       return;
     }
-    string = string.replace(/calc\(/g, '(');
+    string = string.replace(/^calc\(/, '(');
 
     // We tag units by prefixing them with 'U' (note that we are already
     // lowercase) to prevent problems with types which are substrings of
--- a/src/parsing.js
+++ b/src/parsing.js
@@ -1,0 +1,72 @@
+
+// Copyright 2015 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+//     You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(internal, scope, testing) {
+
+  function parseDimension(unitRegExp, string) {
+    string = string.trim().toLowerCase();
+
+    if (string == '0' && 'px'.search(unitRegExp) >= 0)
+      return {px: 0};
+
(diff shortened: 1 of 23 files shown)


nattridgell:
LGTM :+1:

src/parsing.js
```diff
@@ -0,0 +1,83 @@
...
+
+    // CSS allows lengths to be '0', so if px is a supported unit, return 0px.
+    if (string == '0' && 'px'.search(unitRegExp) >= 0)
+      return {px: 0};
+
+    // If we have parenthesis, we're a calc and need to start with 'calc'.
+    if (!/^[^(]*$|^calc/.test(string)) {
+      return;
+    }
+    string = string.replace(/calc\(/g, '(');
```
nattridgell: 'calc(50px + calc(5em))' is invalid.  However calc(50px + (5em)) is valid.  This line makes both expressions the same.

wilddamon: Fixed, and added a test.

src/parsing.js
```diff
@@ -14,14 +14,78 @@
 
 (function(internal, testing) {
 
-  var _numberValueRegex = /^\s*[-+]?(\d*\.)?\d+(e[-+]?\d+)?\s*$/;
+  // Extra backslashes because otherwise JS interprets them incorrectly.
+  var _numberValueRegexStr = '[-+]?(\\d*\\.)?\\d+(e[-+]?\\d+)?';
```
rjwright: Is there a reason this has an underscore in the name? Your knowledge of the style guide leads me to think yes :P

src/parsing.js
```diff
@@ -14,14 +14,78 @@
...
+      return 'U' + match;
+    });
+    var taggedUnitRegExp = 'U(' + unitRegExp.source + ')';
+
+    // Validating input is simply applying as many reductions as we can.
+    var typeCheck = string.replace(new RegExp(_numberValueRegexStr, 'g'), 'N')
+                          .replace(new RegExp('N' + taggedUnitRegExp, 'g'), 'D')
+                          .replace(/\s[+-]\s/g, 'O')
+                          .replace(/\s/g, '');
+    var reductions = [/N\*(D)/g, /(N|D)[*/]N/g, /(N|D)O\1/g, /\((N|D)\)/g];
```
rjwright: Might be good to add a comment about what this is for.

src/parsing.js
```diff
@@ -14,14 +14,78 @@
...
+      } else {
+        i++;
+      }
+    }
+    if (typeCheck != 'D') {
+      return;
+    }
+
+    for (var unit in matchedUnits) {
+      // TODO: Don't use eval here as it will throw up security flags.
```
rjwright, line 94: Do we want to fix this before landing?

wilddamon, line 94: It's non-trivial because this method takes advantage of the behaviour of eval. We'd need to write a cumulating parser for them...

src/parsing.js
wilddamon: Done, let me know if you think this is too verbose.

src/parsing.js
wilddamon: Not really. removed.

rjwright:
LGTM

PR #59, opened 2015-12-22
StylePropertyMapReadOnly get method implemented, parse method implemented for StyleValue and NumberValue

njlawton:


test/js/computed-style-property-map.js
```diff
@@ -8,20 +8,18 @@ suite('Computed StylePropertyMap', function() {
...
-  test('Computed StylePropertyMapReadOnly.get works for properties that can be NumberValues', function() {
+  test('get method returns a null object if the property is not supported', function() {
     var computedStyleMap = getComputedStyleMap(this.element);
-    var opacity = computedStyleMap.get('opacity');
-    assert.instanceOf(opacity, NumberValue);
-    assert.equal(opacity.value, 0.5);
-    assert.equal(opacity.cssString, '0.5');
+    var propertyStyleValue = computedStyleMap.get('lemon');
+
+    assert.equal(propertyStyleValue, null);
```
wilddamon: Let's throw an exception here instead of just returning null.

src/style-value.js
```diff
@@ -16,15 +16,24 @@
 
   function StyleValue() {}
 
-  // TODO: Include parsing logic here.
   StyleValue.parse = function(property, value) {
```
wilddamon, line 20: How will this work if value should become a sequence? Don't need to address this now, just need add a TODO.

test/js/inline-style-property-map.js
```diff
@@ -21,6 +21,16 @@ suite('Inline StylePropertyMap', function() {
     assert.strictEqual(this.element.style['height'], simpleLength.cssString);
   });
 
+  test('The set method successfully sets the CSS string of the list of StyleValues on property ' +
```
wilddamon: This seems unrelated? If you have changes that you would like to make to unrelated parts of the codebase, it's good practice to put them in separate pull requests.

test/js/number-value.js
```diff
@@ -20,4 +20,26 @@ suite('NumberValue', function() {
...
+  });
+
+  test('parse method should return null if string cannot be successfully parsed to a number', function() {
+    assert.strictEqual(NumberValue.parse('hello, world'), null);
+    assert.strictEqual(NumberValue.parse('calc(10px+3.2em) 3px'), null);
+    assert.strictEqual(NumberValue.parse('3px calc(10px+3.2em)'), null);
+    assert.strictEqual(NumberValue.parse('3px'), null);
+    assert.strictEqual(NumberValue.parse('scale(3, -1)'), null);
+    assert.strictEqual(NumberValue.parse('1e-.0'), null);
+    assert.strictEqual(NumberValue.parse('1e4.0'), null);
```
wilddamon, line 44: What about `'6 someOtherRandomString'`?

test/js/property-dictionary.js
```diff
@@ -65,4 +65,15 @@ suite('PropertyDictionary', function() {
 
     assert.throw(function () {cssPropertyDictionary.getListValueSeparator('height')}, TypeError);
   });
+
+  test('The getValidStyleValuesArray should return an array of constructors for all StyleValue types accepted by that propertty', function() {
```
wilddamon: Can leave off "The" at the start.
s/propertty/property (extra t)

'getValidStyleValuesArray should return an array of constructors for all StyleValue types accepted by that property'

src/number-value.js
```diff
@@ -23,7 +23,13 @@
   }
   internal.inherit(NumberValue, internal.StyleValue);
 
+  NumberValue.parse = function(value) {
+    if (/^\s*[-+]?(\d*\.)?\d+(e[-+]?\d+)?\s*$/.test(value)) {
```
wilddamon: Lets move this regex out into a parser file like in my pull request. There will be lots of other properties that will need to parse numbers as part of their parsing, and it makes sense to share this regex with them.

src/style-value.js
```diff
@@ -16,15 +16,24 @@
...
-    if (typeof value == 'string') {
-      numberValue = Number.parseFloat(value);
-      if (numberValue !== NaN) {
-        return new scope.NumberValue(numberValue);
+    if (!propertyDictionary().isSupportedProperty(property)) {
+      return null;
+    }
+
+    if (propertyDictionary().isValidKeyword(property, value)) {
+      return Keyword.parse(value);
```
wilddamon: Isn't this just the same as `new KeywordValue(value)`? It's clearer just to use the constructor in this case.

src/style-property-map-readonly.js
```diff
@@ -19,77 +19,16 @@
...
-      // TODO: Stuff that takes shorthands will need to be handled separately.
-
-      default:
-        throw new TypeError('Not implemented yet');
+    propertyString = this._styleObject[property];
+    if (propertyString == '') {
+      return null;
     }
 
+    return StyleValue.parse(property, propertyString);
```
wilddamon: As discussed, let's add the handling the possibility of arrays coming out of parse now, so

``` javascript
  var parsed = StyleValue.parse(property, propertyString);
  if (typeof parsed == 'Array') {
    return parsed[0];
  }
  return parsed;
```

njlawton:
Not finished yet just needed to clear changes already made

src/style-property-map-readonly.js
```diff
@@ -19,77 +19,20 @@
...
-    if (typeof property != 'string') {
-      throw new TypeError('parameter 1 is not of type \'string\'');
+    if (!propertyDictionary().isSupportedProperty(property)) {
+      throw new TypeError(property + ' is not a supported CSS property');
     }
 
-    var value = this._styleObject[property];
-    if (!value) {
+    propertyString = this._styleObject[property];
+    if (propertyString == '') {
```
wilddamon: You can still write (and I think it's clearer since `'' == false == null` in javascript:

``` javascript
if (!propertyString)
```

PR #61, opened 2015-12-29
CalcLength CssString helper

nattridgell:
- change method name (and make internal method)
- change to '_generateCssString' to fit with the standard
  from other classes.


rjwright:
lgtm

PR #62, opened 2015-12-29
Transform SimpleLength PX type check update

nattridgell:
LengthValue.LengthType enum now updated.
Update checks for type PX as apropriate and remove TODO's.


wilddamon:
LGTM

PR #63, opened 2016-01-04
Implemented the getProperties method in the StylePropertyMap and StylePropertyMapReadOnly class

njlawton:


src/style-property-map.js
```diff
@@ -86,7 +86,23 @@
     if (!propertyDictionary().isSupportedProperty(property)) {
       throw new TypeError('Cannot use has method for ' + property + ' because it is not a supported CSS property');
     }
-    return !!this._styleObject[property]
+    return !!this._styleObject[property];
+  };
+
+  StylePropertyMap.prototype.getProperties = function() {
+    var specifiedPropertyIndex = 0;
+    var isSpecifiedProperty = true;
```
wilddamon: This variable doesn't seem to ever change. If this is intentional it is clearer to omit it and write this below:

``` javascript
while (true) {
  if (this._styleObject[specifiedPropertyIndex] == '') {
```

src/style-property-map.js
```diff
@@ -86,7 +86,23 @@
...
+    var propertyArray = [];
+
+    while (isSpecifiedProperty) {
+      if (this._styleObject[specifiedPropertyIndex] == '') {
```
wilddamon: Can just write
```javascript
if (!this._styleObject[specifiedPropertyIndex]) {
```

src/style-property-map.js
```diff
@@ -86,7 +86,23 @@
...
-    return !!this._styleObject[property]
+    return !!this._styleObject[property];
+  };
+
+  StylePropertyMap.prototype.getProperties = function() {
+    var specifiedPropertyIndex = 0;
+    var isSpecifiedProperty = true;
+    var propertyArray = [];
+
+    while (isSpecifiedProperty) {
```
wilddamon: You're iterating through the properties in the CSSStyleDeclaration (ComputedStyle etc)  here right? If so, CSSStyleDeclaration is iterable, so it has a length and you can write

``` javascript
for (var i = 0; i < this._styleObject.length; i++) {
  propertyArray.push(this._styleObject[i]);
}
```

src/style-property-map.js
```diff
@@ -86,7 +86,23 @@
...
+      if (this._styleObject[specifiedPropertyIndex] == '') {
+        break;
+      }
+      propertyArray[specifiedPropertyIndex] =
+        this._styleObject[specifiedPropertyIndex];
+      specifiedPropertyIndex++;
+    }
+    return propertyArray.sort();
```
wilddamon: Does the spec require this to be sorted? If not, let's not bother sorting the properties.

njlawton:
Done

src/style-property-map-readonly.js
```diff
@@ -36,13 +36,12 @@
...
   StylePropertyMapReadOnly.prototype.getProperties = function() {
-    var output = [];
-    for (var i = 0, l = this._styleObject.length; i < l; ++i) {
-      var property = this._styleObject[i];
-      // TODO: Construct the objects for each type of thing and add them to the
-      // output array.
+    var cssPropertiesFullText = this._styleObject.cssText.split('; ');
+    var cssProperties = [];
+    for (var i = 0; i < cssPropertiesFullText.length; i++) {
+      cssProperties[i] = cssPropertiesFullText[i].split(':')[0];
```
wilddamon: Sorry I didn't notice this before. It'd be more efficient to avoid parsing this string and doing what you did in `StylePropertyMap.prototype.getProperties`.

Also, if you move that getProperties to here, you don't need the one in StylePropertyMap since it inherits from this one.

wilddamon: Sorry I didn't notice this earlier - it's generally better to avoid parsing strings if you can, and you should only need to implement the getProperties in StylePropertyMapReadonly since StylePropertyMap inherits from it. If you move the logic from StylePropertyMap.prototype.getProperties to here, then everything should work as intended.

``` javascript
>>> a = getComputedStyle($0)
CSSStyleDeclaration {0: "animation-delay", ... }
>>> a.length
264
>>> for (var i = 0; i < a.length; i++) { console.log(a[i]); }
animation-delay
animation-direction
animation-duration
...

>>> b = $0.style
CSSStyleDeclaration {alignContent: "", ...}
>>> b.length
0
>>> for (var i = 0; i < b.length; i++) { console.log(b[i]); }
undefined
```

njlawton:
StylePropertyMapReadOnly getProperties method has been fixed

wilddamon:
LGTM

PR #65, opened 2016-01-06
StylePropertyMap: Updated the get method to work with properties that contain a sequence of StyleValues

njlawton:


src/style-property-map-readonly.js
```diff
@@ -29,7 +29,7 @@
     }
 
     var parsed = StyleValue.parse(property, propertyString);
-    if (typeof parsed == 'Array') {
+    if (parsed.constructor == Array) {
```
wilddamon, line 32: Out of curiosity, why this change?

src/style-value.js
```diff
@@ -16,34 +16,55 @@
 
   function StyleValue() {}
 
-  // TODO: Add support for cssString being a sequence.
-  StyleValue.parse = function(property, cssString) {
-    if (typeof property != 'string') {
-      throw new TypeError('Property name must be a string');
-    }
-    if (typeof cssString != 'string') {
-      throw new TypeError('Must parse a string');
```
wilddamon: You seem to have lost some of my changes.

src/style-value.js
```diff
@@ -16,34 +16,55 @@
...
+    //Currently only supports sequences separated by ', '
+    var valueArray = value.split(', ');
+    var styleValueArray = [];
+    var supportedStyleValues =
+        propertyDictionary().getValidStyleValuesArray(property);
+
+    for (var i = 0; i < valueArray.length; i++) {
+      var cssString = valueArray[i];
+      cssString = cssString.trim().toLowerCase();
+      if (typeof cssString != 'string') {
```
wilddamon: The check that value is a string must be before you call split - otherwise you will get an exception on line 29 before you get to this line.

src/style-value.js
```diff
@@ -16,34 +16,55 @@
 
   function StyleValue() {}
 
-  // TODO: Add support for cssString being a sequence.
-  StyleValue.parse = function(property, cssString) {
```
wilddamon, line 20: I preferred calling this cssString rather than value here. Please put that back :)

src/style-value.js
```diff
@@ -16,34 +16,55 @@
...
-    if (typeof property != 'string') {
-      throw new TypeError('Property name must be a string');
-    }
-    if (typeof cssString != 'string') {
```
wilddamon, line 24: You need to keep this check here.

src/style-value.js
```diff
@@ -16,34 +16,55 @@
...
+      for (var j = 0; j < supportedStyleValues.length; j++) {
+        styleValueObject = supportedStyleValues[j].parse(cssString);
+        if (styleValueObject != null) {
+          styleValueArray[i] = styleValueObject;
+          successfulParse = true;
+          break;
+        }
+      }
+
+      if (successfulParse == false) {
```
wilddamon:
``` javascript
if (!successfulParse) {
```

src/style-value.js
```diff
@@ -16,34 +16,55 @@
...
+      }
 
-    var styleValueObject = null;
-    var supportedStyleValues = propertyDictionary().getValidStyleValuesArray(property);
-    for (var i = 0; i < supportedStyleValues.length; i++) {
-      styleValueObject = supportedStyleValues[i].parse(cssString);
-      if (styleValueObject != null) {
-        return styleValueObject;
+      var styleValueObject = null;
+      var successfulParse = false;
```
wilddamon: You can define this outside the loop to avoid initialising it many times.

src/style-value.js
```diff
@@ -16,34 +16,55 @@
...
+        throw new TypeError(property +
+          ' has an unsupported StyleValue type or Sequence value separator');
       }
     }
-    return styleValueObject;
+    return styleValueArray;
+  };
+
+  StyleValue.prototype.
+      _getValueArray = function(property, value) {
```
wilddamon: This function doesn't seem to be used?

test/js/computed-style-property-map.js
```diff
@@ -23,6 +23,18 @@ suite('Computed StylePropertyMap', function() {
...
+    'of StyleValues', function() {
+    var inlineStyleMap = this.element.styleMap();
+    var computedStyleMap = getComputedStyleMap(this.element);
+    var valueArray = [new NumberValue(4), new NumberValue(5), new KeywordValue('infinite')];
+    inlineStyleMap.set('animation-iteration-count', valueArray);
+    var propertyStyleValue = computedStyleMap.get('animation-iteration-count');
+
+    assert.instanceOf(propertyStyleValue, NumberValue);
+    assert.strictEqual(propertyStyleValue.cssString, '4');
+  });
```
wilddamon, line 36: It would also be great to have a test for the parse method directly - you can create a new test file for StyleValue.

src/style-property-map-readonly.js
njlawton, line 32: When looking through JavaScript resources this was suggested as a more common way of doing this. This is no big deal either way and I will change it if you would like me to.

test/js/computed-style-property-map.js
njlawton, line 36: Done

src/style-property-map-readonly.js
wilddamon, line 32: Best to keep it consistent with everything else we have.

src/style-value.js
```diff
@@ -29,21 +28,44 @@
       throw new TypeError('Can\'t parse an unsupported property.');
     }
 
-    // Make sure that there is no leading or trailing whitespace, case insensitive.
-    cssString = cssString.trim().toLowerCase();
-    if (propertyDictionary().isValidKeyword(property, cssString)) {
-      return new KeywordValue(cssString);
-    }
+    //Currently only supports sequences separated by ', '
```
wilddamon: Nitpick: needs a space after //

src/style-value.js
```diff
@@ -29,21 +28,44 @@
...
     var styleValueObject = null;
-    var supportedStyleValues = propertyDictionary().getValidStyleValuesArray(property);
-    for (var i = 0; i < supportedStyleValues.length; i++) {
-      styleValueObject = supportedStyleValues[i].parse(cssString);
-      if (styleValueObject != null) {
-        return styleValueObject;
+    var successfulParse = false;
+    for (var i = 0; i < valueArray.length; i++) {
+      var cssStringStyleValue = valueArray[i];
+      cssStringStyleValue = cssStringStyleValue.trim().toLowerCase();
```
wilddamon: I think it'd be better to do the toLowerCase() on the whole cssString before you split it. Then you only need to trim here.

src/style-value.js
```diff
@@ -29,21 +28,44 @@
...
+        continue;
+      }
+
+      styleValueObject = null;
+      successfulParse = false;
+      for (var j = 0; j < supportedStyleValues.length; j++) {
+        try {
+          styleValueObject = supportedStyleValues[j].parse(cssStringStyleValue);
+        } catch (e) {
+          //Ensures method does not terminate if a StyleValue parse method throws an error
```
wilddamon: Same nitpick as earlier in the file. Needs space after //

wilddamon:
LGTM once nitpicks are fixed.

src/style-property-map-readonly.js
njlawton, line 32: As it turns out you cant use typeof to refer to arrays as the function will return 'object'

wilddamon:
LGTM

PR #66, opened 2016-01-12
StylePropertyMapReadOnly getAll method implemented

njlawton:
Also used the getAll method to simplify the implementation of the get method in StylePropertyMapReadOnly


wilddamon:
lgtm

PR #67, opened 2016-01-13
Use createElement() to load scripts

surma:
The current approach to loading the scripts is both inefficient and prone to breaking.
- `document.write` is notoriously slow and a generally discouraged API. We can avoid starting the HTML parser by assembling the DOM node ourselves.
- The current script relies on the `typed-om.js` include to be the very last include. It would be more common to provide a global variable to configure the import path (admittedly, still not the best way, but the only one that works without bigger changes).
- Globals should be accessed with the `window` object.

### DIFF
--- a/target-loader.js
+++ b/target-loader.js
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 (function() {
-  var scripts = document.getElementsByTagName('script');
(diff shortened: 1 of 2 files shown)


wilddamon:
This seems fine to me, just waiting for Renee who has done more of this kind of work than I have.

rjwright:
LGTM.

PR #68, opened 2016-01-13
Use npm script to run tests in headless browser

surma:
This is only one of many steps necessary to make this a proper packages (usable by tools like `bower` and the like). But having `sudo npm install XXX` in the instructions is definitely not a good idea and is also unnecessary. Secondly, having to open the browser to run the tests is definitely going to make a few devs avoid it :wink: 

This change turns the whole repository into an npm package, even though not published at this point. That is your prerogative. Please take a close look if the info I used in `package.json` looks good to you. I added a simple npm directive to make `npm test` work, which runs the test in [PhantomJS](http://phantomjs.org/) and reports on the command line.

screenshot 2016-01-13 14 47 43

Yes, some tests are failing right now as, sadly, PhantomJS is rather old. But this was more about the unit tests. The old open-in-a-browser approach still works.

Further work can be done here once #67 has landed or been veto’d.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "typed-om",
+  "version": "1.0.0",
+  "description": "Prototype for CSS Typed OM",
+  "directories": {
+    "test": "test"
+  },
+  "devDependencies": {
+    "chai": "^3.4.1",
+    "mocha": "^2.3.4",
+    "mocha-phantomjs": "^4.0.2"
+  },
+  "scripts": {
+    "test": "./node_modules/mocha-phantomjs/bin/mocha-phantomjs ./test/runner.html"
+  },
+  "repository": {
+    "type": "git",
+    "url": "github.com/css-typed-om/typed-om"
+  },
+  "license": "Apache-2.0"
+}
--- a/package.json
+++ b/package.json
@@ -8,10 +8,11 @@
   "devDependencies": {
     "chai": "^3.4.1",
     "mocha": "^2.3.4",
-    "mocha-phantomjs": "^4.0.2"
+    "mocha-phantomjs-core": "^1.3.0",
+    "phantomjs": "^2.1.2"
   },
   "scripts": {
-    "test": "./node_modules/mocha-phantomjs/bin/mocha-phantomjs ./test/runner.html"
+    "test": "./node_modules/phantomjs/bin/phantomjs ./node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js ./test/runner.html"
   },
   "repository": {
     "type": "git",
--- a/test/js/computed-style-property-map.js
+++ b/test/js/computed-style-property-map.js
@@ -23,7 +23,7 @@ suite('Computed StylePropertyMap', function() {
     assert.strictEqual(propertyStyleValue.cssString, '0.5');
   });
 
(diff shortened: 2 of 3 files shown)


rjwright:
So some of our tests are failing when you run them this way?

surma:
Yeah, I should have put that at the top of my post (it’s under the screenshot):

> Yes, some tests are failing right now as, sadly, PhantomJS is rather old. But this was more about the unit tests. The old open-in-a-browser approach still works.

I will try to come up with a workaround.

rjwright:
Yeah I just wasn't 100% sure which tests you were referring to.

A workaround would be good. This polyfill is still a proof of concept, so we don't expect many external devs to be working on it or even using it. Your change is probably still worthwhile for our own convenience, but I would prefer if all the tests were passing first.

surma:
So, for the [second failing test](https://github.com/css-typed-om/typed-om/blob/master/test/js/computed-style-property-map.js#L31) it seems like PhantomJS auto-prefixes the `animation-iteration-count` with `-webkit-`. `propertyStyleValue` is therefore `null`. Any other value we can add to the `PropertyDictionary` and use instead? I was thinking `font-face` but I don’t think we have implemented something like `StringValue` (I know that that is not in the standard, but we have to handle these values, right?). @rjwright WDYT?

Still trying to figure out why the [first failing test](https://github.com/css-typed-om/typed-om/blob/master/test/js/length-value.js#L79) is failing. My guess is that PhantomJS has some problems with the RegExps (maybe backreferences?).

surma:
Actually, `StringValue` is in the [spec](https://drafts.css-houdini.org/css-typed-om/#mapping-of-properties-to-accepted-types) (as is `TimeValue` and a few others), just not properly defined at this point.

surma:
Ha! Talking to actual engineers seems to have refreshed my mind.

Made it work without big hacks or inconveniences. I did however disable the test for sequences in `StylePropertyMap` for now.

package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "typed-om",
```
rjwright, line 2: Please use single quotes, if possible.

surma, line 2: For JS files I agree, but in a JSON file that’s strictly speaking invalid as the spec demands double quotes and `npm` actually fails parsing with single quotes.

test/js/computed-style-property-map.js
```diff
@@ -23,7 +23,7 @@ suite('Computed StylePropertyMap', function() {
     assert.strictEqual(propertyStyleValue.cssString, '0.5');
   });
 
-  test('get method returns the first StyleValue in the sequence if a property has been set a sequence ' +
+  test.skip('get method returns the first StyleValue in the sequence if a property has been set a sequence ' +
```
rjwright, line 30: Add a TODO about replacing animation-iteration-count with some other sequence-compatible property & enabling this test.

package.json
rjwright, line 2: Cool. I am not really fluent in JSON. Thanks :)

test/js/computed-style-property-map.js
surma, line 30: Forgot about that. Fixed.

rjwright:
LGTM

wilddamon:
LGTM

PR #69, opened 2016-01-18
Resolve unknown references

surma:
Both `LengthValue` and `StyleValue` have only been attached to the `internal` object but were used here and there without the `internal` object. However, according to spec, both Types are actually exposed as they have the `parse()` methods attached to them. 

I exposed these 2 types again.

### DIFF
--- a/src/property-dictionary.js
+++ b/src/property-dictionary.js
@@ -12,7 +12,7 @@
 //     See the License for the specific language governing permissions and
 // limitations under the License.
 
-(function(internal, scope, testing) {
+(function(internal, testing) {
 
   function PropertyDictionary() {
     this._validProperties = {
--- a/src/property-dictionary.js
+++ b/src/property-dictionary.js
@@ -12,7 +12,7 @@
 //     See the License for the specific language governing permissions and
 // limitations under the License.
 
-(function(shared, scope, testing) {
+(function(internal, scope, testing) {
 
   function PropertyDictionary() {
     this._validProperties = {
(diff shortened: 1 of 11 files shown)


wilddamon:
This mostly LGTM, however I'm wondering if there's a usual way available to disallow things like

``` javascript
new LengthValue()
```

while still allowing this?

``` javascript
LengthValue.parse(...)
```

surma:
AFAIK, private constructors are not really a thing :confused: 
However, the constructor behaves like a copy constructor and errors if something invalid is passed in.  So I think we are actually pretty well-behaved.

src/style-value.js
```diff
@@ -68,7 +68,7 @@
     return styleValueArray;
   };
 
-  internal.StyleValue = StyleValue;
+  scope.StyleValue = StyleValue;
```
wilddamon, line 73: If we're exposing this here, we should probably throw an exception if someone tries to write `new StyleValue()` too.

src/property-dictionary.js
```diff
@@ -12,7 +12,7 @@
 //     See the License for the specific language governing permissions and
 // limitations under the License.
 
-(function(shared, scope, testing) {
+(function(internal, scope, testing) {
```
wilddamon: While we're here, `scope` is unused in this file.

wilddamon:
Cool ok! Minor comments, other than that LGTM

surma:
Agreed and implemented :)

wilddamon:
LGTM

PR #70, opened 2016-01-28
Added support for transform css property in PropertyDictionary

njlawton:
A related bug was also fixed when supporting transform; the cssString for Skew was incorrect as it was missing 'deg' after each number e.g. skew(2, 2) instead of skew(2deg, 2deg) this issue has been fixed.

Note: Due to limitations of the StyleValue parsing method in the polyfill the get and getAll methods in StylePropertyMapReadOnly will not work for transform and will throw an error.

### DIFF
--- a/src/property-dictionary.js
+++ b/src/property-dictionary.js
@@ -20,7 +20,8 @@
      'pitch-range': [NumberValue],
      'border-top-width': [LengthValue],
      'opacity': [NumberValue],
-     'animation-iteration-count': [NumberValue]
+     'animation-iteration-count': [NumberValue],
+     'transform': [TransformValue]
     };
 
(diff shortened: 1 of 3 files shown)


src/property-dictionary.js
```diff
@@ -20,7 +20,8 @@
      'pitch-range': [NumberValue],
      'border-top-width': [LengthValue],
      'opacity': [NumberValue],
-     'animation-iteration-count': [NumberValue]
+     'animation-iteration-count': [NumberValue],
+     'transform': [TransformValue]
```
wilddamon: It'd be good to get these in alphabetical order sooner rather than later. Same below.

rjwright:
LGTM

PR #71, opened 2016-01-28
Make loader maintain script order

surma:
This fix makes the loader loads scripts in the order they are specified in the respective arrays, waiting for one script to finish before loading the next. It makes the entire thing pretty slow again, but this will all go away anyways if we concat all the scripts for “production” (if we ever want to got there :wink: )

r: @rjwright

### DIFF
--- a/target-loader.js
+++ b/target-loader.js
@@ -13,9 +13,20 @@
 // limitations under the License.
 
 (function() {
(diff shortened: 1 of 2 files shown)


wilddamon:
So I think we should move the loadScript function into one file, but lets do that in a separate PR so that Nick gets unblocked. Otherwise LGTM.

surma:
Will do :)

PR #73, opened 2016-01-29
Implemented empty constructor for TransformValue

njlawton:
Updated and added tests to validate the behavior of the constructor.

Current behavior
-Empty constructor and empty array in constructor produce the same object
-Empty constructor asMatrix method will give the 2D identity matrix as output
-Empty constructor cssString will give an empty string as output

Note: As of now the final behavior of an empty TransformValue has not been added to the typed om spec and so may need to change when this has been better defined.

### DIFF
--- a/src/transform-value.js
+++ b/src/transform-value.js
@@ -15,28 +15,12 @@
 (function(internal, scope) {
 
   function TransformValue(values) {
-    if (!Array.isArray(values)) {
-      throw new TypeError('TransformValue must be an array of ' +
-          'TransformComponents.');
-    } else if (values.length < 1) {
-      throw new TypeError('TransformValue must have at least 1 ' +
-          'TransformComponent.');
+    if (values === undefined) {
+      return this._createEmptyTransform();
     }
-
-    this.transformComponents = [];
-    for (var i = 0; i < values.length; i++) {
-      if (!(values[i] instanceof TransformComponent)) {
-        throw new TypeError('Argument at index ' + i + ' is not an instance ' +
-            'of \'TransformComponent\'.');
-      }
-      this.transformComponents.push(values[i]);
-    }
-
-    this._matrix = this._computeMatrix();
-    this.cssString = this._generateCssString();
+    return this._createSequenceTransform(values);
   }
   internal.inherit(TransformValue, StyleValue);
-
 
   TransformValue.prototype.asMatrix = function() {
     return this._matrix;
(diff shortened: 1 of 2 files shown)


src/transform-value.js
```diff
@@ -15,29 +15,13 @@
...
-      if (!(values[i] instanceof TransformComponent)) {
-        throw new TypeError('Argument at index ' + i + ' is not an instance ' +
-            'of \'TransformComponent\'.');
-      }
-      this.transformComponents.push(values[i]);
-    }
-
-    this._matrix = this._computeMatrix();
-    this.cssString = this._generateCssString();
+    return this._createSequenceTransform(values);
```
wilddamon: These functions don't return anything, so this looks a bit strange. Also there's a lot of indirection in the constructor that jumps all around the file now. It's also strange that no argument is treated differently to passing in an empty sequence. Maybe @shans should clarify what the expected behaviour for these cases is? It is not currently covered in the spec.

It would be clearer to keep the setting of _matrix and cssString in the constructor, but  modify the computeMatrix function to look like this:

``` javascript
  function TransformValue(values) {
    if (values === undefined) {
      values = [];
    }
    if (!Array.isArray(values)) {
      throw ...
    }
// Keep the old version of the constructor code
  }
...

  TransformValue.prototype._computeMatrix = function() {
    if (!this.transformComponents.length) {
      // return identity matrix
    }
    var matrix = this.transformComponents[0].asMatrix();
    for (var i = 1; i < this.transformComponents.length; ++i) {
      matrix = matrix.multiply(this.transformComponents[i].asMatrix());
    }
    return matrix;
  };
```

src/transform-value.js
```diff
@@ -61,6 +45,30 @@
     return this.transformComponents.map(getCssString).join(' ');
   };
 
+  TransformValue.prototype._createEmptyTransform = function() {
+    this._matrix = new Matrix(1, 1, 1, 1, 1, 1);
```
wilddamon: The identity matrix is 

``` javascript
new Matrix(1, 0, 0, 1, 0, 0);
```

test/js/transform-value.js
```diff
@@ -9,14 +9,21 @@ suite('TransformValue', function() {
...
+    assert.throws(function() {new TransformValue(null)});
     assert.throws(function() {new TransformValue({})});
-    assert.throws(function() {new TransformValue([])});
     assert.throws(function() {new TransformValue(['1', '2'])});
     assert.throws(function() {new TransformValue([null])});
     assert.throws(function() {new TransformValue([new NumberValue(5)])});
   });
 
+  test('TransformValue empty constructor should create an object with ' + 
+    'the expected properties', function() {
```
rjwright: Maybe say that it should represent a 2D identity matrix with an empty cssString?

rjwright:
Do we need this functionality?

LGTM w/ nit.

PR #76, opened 2016-02-10
Changed ColorValues cssString behaviour to match chrome implementation

njlawton:
Now if alpha is set to 1 either explicitly or implicitly the rgb(..) cssString will be returned


wilddamon:
lgtm

PR #80, opened 2016-05-02
Rename 'parse', 'fromDictionary' and 'fromValue' to 'from'

rjwright:
In the spec this change is specific to LengthValue, but in the code some (not sure why not all) of the other StyleValue "subclasses" also implement parse, which is used by StyleValue.parse. Should I rename 'parse' to 'from' in those subclasses too?

### DIFF
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -97,8 +97,6 @@
     }
   };
 
-  CSSLengthValue.parse = CSSLengthValue.from;
-
   CSSLengthValue.prototype.add = function(addedLength) {
     if (this instanceof CSSSimpleLength &&
         addedLength instanceof CSSSimpleLength &&
--- a/src/length-value.js
+++ b/src/length-value.js
@@ -67,15 +67,20 @@
     }
   };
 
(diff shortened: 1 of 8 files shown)


src/length-value.js
```diff
@@ -67,15 +67,20 @@
     }
   };
 
-  LengthValue.parse = function(cssString) {
-    if (typeof cssString != 'string') {
-      throw new TypeError('Must parse a length out of a string.');
+  LengthValue.from = function(value, type) {
+    if (type !== undefined) {
+      return new SimpleLength(value, type);
+    } else if (typeof value == 'object') {
```
wilddamon: This can just be

``` javascript
if (...) {
  return new SimpleLength(...);
}
if (...) {
  return ...
}
if (...) {
...
```

wilddamon:
I wouldn't bother renaming the parse methods. They exist in some places and not others because we didn't get round to it I think?

src/length-value.js
rjwright: Done

PR #81, opened 2016-05-02
Prefix StyleValue and its subclasses with "CSS"

rjwright:
StyleValue -> CSSStyleValue
CalcLength -> CSSCalcLength
ColorValue -> CSSColorValue
KeywordValue -> CSSKeywordValue
LengthValue -> CSSLengthValue
NumberValue -> CSSNumberValue
PositionValue -> CSSPositionValue
SimpleLength -> CSSSimpleLength
TransformValue -> CSSTransformValue

Do I need to rename the files?

### DIFF
--- a/src/calc-length.js
+++ b/src/calc-length.js
@@ -14,14 +14,14 @@
 
 (function(internal, scope) {
 
-  // TODO: CalcLength(cssString)
-  function CalcLength(dictionary) {
+  // TODO: CSSCalcLength(cssString)
+  function CSSCalcLength(dictionary) {
     if (typeof dictionary != 'object') {
-      throw new TypeError('CalcLength must be passed a dictionary object');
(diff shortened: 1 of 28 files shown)


wilddamon:
It's probably better to rename the files as well. Maybe do it in a new PR for clarity?

rjwright:
> It's probably better to rename the files as well. Maybe do it in a new PR for clarity?

Sounds good. Can we merge this one so I can do the transform components prefixing? Then I will rename files last.

wilddamon:
Actually, I think it'd be clearer to do it class-by-class in the future (including the file renames), but sure.

rjwright:
> Actually, I think it'd be clearer to do it class-by-class in the future (including the file renames), but sure.

Opps. I'm already almost finished the transformation stuff too. Oh well, good to know for the Blink change.

Thanks.

PR #90, opened 2016-06-29
Add CSSAngleValue and a test

wilddamon:
Spec: https://drafts.css-houdini.org/css-typed-om/#anglevalue-objects

### DIFF
--- a/src/css-angle-value.js
+++ b/src/css-angle-value.js
@@ -21,28 +21,28 @@
     TURN: "turn"
   };
 
-  function isValidAngleUnit(unit) {
+  function isValidAngleType(unit) {
     return internal.objects.any(AngleType, function(type) {
       return unit == type;
     });
   }
 
-  function rad2deg(rad) {
+  function radToDeg(rad) {
     return rad * 180 / Math.PI;
   }
-  function grad2deg(grad) {
+  function gradToDeg(grad) {
     return grad * 9 / 10;
   }
-  function turn2deg(turn) {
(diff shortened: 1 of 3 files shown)


src/css-angle-value.js
```diff
@@ -0,0 +1,95 @@
...
+(function(internal, scope) {
+
+  AngleType = {
+    DEG: "deg",
+    RAD: "rad",
+    GRAD: "grad",
+    TURN: "turn"
+  };
+
+  function isValidAngleUnit(unit) {
```
rjwright: This is analogous to CSSLengthValue.isValidLengthType. Should make terminology consistent.

src/css-angle-value.js
```diff
@@ -0,0 +1,95 @@
...
+    return internal.objects.any(AngleType, function(type) {
+      return unit == type;
+    });
+  }
+
+  function rad2deg(rad) {
```
rjwright: I would prefer to call these radToDeg etc. unless there is some precedent for this naming that I don't know about.

src/css-angle-value.js
wilddamon: Changed to isValidAngleType. Kept it as a private function for now.

src/css-angle-value.js
wilddamon: Done. These are what the internal methods in blink are called. No need to follow that though.

rjwright:
lgtm

PR #91, opened 2016-06-29
Rename cssString -> cssText

wilddamon:
This is an automated rename of cssString to cssText. It was done using

grep -rl cssString . | xargs sed -e -i 's/cssString/cssText/'

I looked over it, and didn't see anything that got renamed that shouldn't have.

Corresponding spec change: https://github.com/w3c/css-houdini-drafts/pull/108

PTAL.

### DIFF
--- a/src/css-calc-length.js
+++ b/src/css-calc-length.js
@@ -14,7 +14,7 @@
 
 (function(internal, scope) {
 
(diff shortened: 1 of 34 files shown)


rjwright:
lgtm

PR #93, opened 2016-06-29
Rename is2DComponent to is2D.

wilddamon:
This is an automated rename of CSSTransformValue.is2DComponent and CSSTransformComponent.is2DComponent to just is2D.

The command used was

grep -rl is2DComponent . | xargs sed -i -e 's/is2DComponent/is2D/'

See the spec here: https://drafts.css-houdini.org/css-typed-om/#transformvalue-objects, https://drafts.css-houdini.org/css-typed-om/#csstransformcomponent

### DIFF
--- a/src/css-matrix.js
+++ b/src/css-matrix.js
@@ -75,7 +75,7 @@
     return this;
   };
 
-  CSSMatrix.prototype.is2DComponent = function() {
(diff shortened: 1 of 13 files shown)


rjwright:
lgtm

PR #94, opened 2016-06-29
Add parsing for CSSColorValue

wilddamon:
Note: This will conflict with https://github.com/css-typed-om/typed-om/pull/91, if you merge that one first this will need updating.

### DIFF
--- a/src/parsing.js
+++ b/src/parsing.js
@@ -113,16 +113,16 @@
 
   // Regex should be anchored with /^
   function consumeToken(regex, string) {
-    var result = regex.exec(string);
-    if (result) {
-      result = regex.ignoreCase ? result[0].toLowerCase() : result[0];
-      return [result, string.substr(result.length)];
+    var match = regex.exec(string);
+    if (match) {
+      match[0] = regex.ignoreCase ? match[0].toLowerCase() : match[0];
+      return [match[0], string.substr(match[0].length)];
     }
   }
 
   function consumeNumber(string) {
     var result = consumeToken(/^[0-9]?\.?[0-9]+/, string);
-    if (result && result[0]) {
+    if (result) {
       result[0] = parseFloat(result[0]);
     }
     return result;
@@ -132,7 +132,9 @@
     string = string.replace(/^\s*/, '');
     var result = consumer(string);
     if (result) {
-      return [result[0], result[1].replace(/^\s*/, '')];
+      // Remove whitespace from the start of the remainder string too.
+      result[1] = result[1].replace(replace(/^\s*/, '');
+      return result;
     }
   }
 
@@ -169,7 +171,6 @@
     }
   }
 
-  internal.parsing = {};
   internal.parsing.isNumberValueString = isNumberValueString;
   internal.parsing.isCalc = isCalc;
   internal.parsing.parseDimension = parseDimension;
--- a/src/parsing.js
+++ b/src/parsing.js
@@ -12,7 +12,7 @@
 //     See the License for the specific language governing permissions and
 // limitations under the License.
 
-(function(internal, testing) {
+(function(internal) {
 
   // Extra backslashes because otherwise JS interprets them incorrectly.
   var numberValueRegexStr = '[-+]?(\\d*\\.)?\\d+(e[-+]?\\d+)?';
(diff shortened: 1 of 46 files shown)


src/parsing.js
```diff
@@ -102,10 +102,83 @@
...
+    var result = regex.exec(string);
+    if (result) {
+      result = regex.ignoreCase ? result[0].toLowerCase() : result[0];
+      return [result, string.substr(result.length)];
+    }
+  }
+
+  function consumeNumber(string) {
+    var result = consumeToken(/^[0-9]?\.?[0-9]+/, string);
+    if (result && result[0]) {
```
rjwright: If there is result there must be result[0], right?

src/parsing.js
```diff
@@ -102,10 +102,83 @@
...
+      var result = value(input);
+      if (result)
+        result[0] = undefined;
+      return result;
+    }
+  }
+
+  // Regex should be anchored with /^
+  function consumeToken(regex, string) {
+    var result = regex.exec(string);
```
rjwright: This & the other consume methods were a bit confusing to me. I think it would be clearer if we renamed "result" to "matched" or similar here (but keep using "result" for the pairs below).

src/parsing.js
```diff
@@ -102,10 +102,83 @@
...
+    if (result) {
+      result = regex.ignoreCase ? result[0].toLowerCase() : result[0];
+      return [result, string.substr(result.length)];
```
rjwright: Would you be opposed to changing this to something like
`return { matched: result, remainder: string.substr(result.length) }`

src/parsing.js
```diff
@@ -102,10 +102,83 @@
...
+      result[0] = parseFloat(result[0]);
+    }
+    return result;
+  }
+
+  function consumeTrimmed(consumer, string) {
+    string = string.replace(/^\s*/, '');
+    var result = consumer(string);
+    if (result) {
+      return [result[0], result[1].replace(/^\s*/, '')];
```
rjwright: Nit
`if (result) {
    result[1] = result[1].replace(/^\s*/, '');
}
return result;
`
to match consumeNumber?

test/js/css-matrix.js
```diff
@@ -20,8 +20,7 @@ suite('CSSMatrix', function() {
       value = new CSSMatrix(10, 20, -0.5, 0.5, 4, 2);
     });
     assert.isTrue(value.is2D());
-    assert.strictEqual(value.cssText,
-      'matrix(10, 20, -0.5, 0.5, 4, 2)');
+    assert.strictEqual(value.cssText, 'matrix(10, 20, -0.5, 0.5, 4, 2)');
```
rjwright: Are these changes meant to be in this PR?

rjwright:
Looking great, but I want to have another quick look at it tomorrow.

test/js/css-color-value.js
```diff
@@ -46,4 +46,60 @@ suite('CSSColorValue', function() {
...
+    var color = CSSColorValue.from('rgba(45,12,56,0.5)');
+
+    assert.instanceOf(color, CSSColorValue);
+    assert.strictEqual(color.r, 45);
+    assert.strictEqual(color.g, 12);
+    assert.strictEqual(color.b, 56);
+    assert.strictEqual(color.a, 0.5);
+  });
+
+  test('Whitespace is ignore when parsing', function() {
```
rjwright: s/ignore/ignored

rjwright:
lgtm with comments

src/parsing.js
wilddamon: I think that'll make the code pretty verbose, but not strongly opposed? What if I just add a comment saying that the parseX functions all return [consumed, remainder] at the top?

src/parsing.js
wilddamon: Done

src/parsing.js
wilddamon: I think I put that there because I ran into a case where there wasn't while I was developing this. However in theory there shouldn't be now that I finished consumeToken (I hope!), so I've removed it :)

src/parsing.js
wilddamon: Done.

test/js/css-color-value.js
wilddamon: Done.

test/js/css-matrix.js
wilddamon: No, removed.

src/parsing.js
rjwright: Yeah my idea was either we change this to an object, or add a comment. I always lean toward making the code more expressive first, but if you would prefer a comment that's fine.

rjwright:
LGTM

PR #95, opened 2016-06-30
Make semi-private methods in LengthValues properly hidden

wilddamon:
Here I've used scoping to fully hide most methods we don't want developers to see. Some were needed in LengthValue (simpleLengthToCalcLength, addSimpleLengths, subtractSimpleLengths), so I put them in in the internal scope. I think this makes LengthValue's add and subtract methods much tidier - WDYT?

### DIFF
--- a/src/css-calc-length.js
+++ b/src/css-calc-length.js
@@ -14,7 +14,6 @@
 
 (function(internal, scope) {
 
-  // TODO: CSSCalcLength(cssString)
   function CSSCalcLength(dictionary) {
     if (typeof dictionary != 'object') {
       throw new TypeError('CSSCalcLength must be passed a dictionary object');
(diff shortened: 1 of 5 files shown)


src/css-simple-length.js
```diff
@@ -39,29 +39,29 @@
...
-    if (this.type != subtractedLength.type) {
+    if (length.type != subtractedLength.type) {
       throw new TypeError('CSSSimpleLength units are not the same');
     }
-    return new CSSSimpleLength((this.value - subtractedLength.value), this.type);
+    return new CSSSimpleLength(length.value - subtractedLength.value, length.type);
   };
 
-  CSSSimpleLength.prototype._asCalcLength = function() {
+  function toCalcLength(simpleLength) {
```
rjwright: Can we rename this to simpleLengthToCalcLength?

src/css-calc-length.js
```diff
@@ -104,47 +103,39 @@
 
     // Iterate through all possible length types and add their values.
     internal.objects.foreach(CSSLengthValue.LengthType, function(type) {
-      if (this[type] == null && addedLength[type] == null) {
+      if (calcLength[type] == null && addedLength[type] == null) {
```
rjwright: Can this not be simplified in the same way as subtractCalcLengths?

rjwright:
lgtm with comments

src/css-simple-length.js
wilddamon: Done.

src/css-calc-length.js
wilddamon: It can, done.

PR #96, opened 2016-06-30
Move Matrix code into a new class DOMMatrixReadonly

wilddamon:
This changes CSSMatrix to behave more like the spec. 
DOMMatrixReadonly isn't generally available in Chrome yet, so this can act as a shim for that.

There's a bunch of horrible asMatrix().matrix calls everywhere, I hope to get rid of that in a later CL where we remove the asMatrix() method and replace it with a DOMMatrixReadonly property called matrix.

I also wasn't sure about where to put the testing method matricesApproxEqual. Do you know if there's a usual way to handle that?


src/dom-matrix-readonly.js
```diff
@@ -0,0 +1,170 @@
...
+    assert.closeTo(matrix1.m33, matrix2.m33, EPSILON);
+    assert.closeTo(matrix1.m34, matrix2.m34, EPSILON);
+    assert.closeTo(matrix1.m41, matrix2.m41, EPSILON);
+    assert.closeTo(matrix1.m42, matrix2.m42, EPSILON);
+    assert.closeTo(matrix1.m43, matrix2.m43, EPSILON);
+    assert.closeTo(matrix1.m44, matrix2.m44, EPSILON);
+  }
+
+  internal.DOMMatrixReadonly = DOMMatrixReadonly;
+  if (TYPED_OM_TESTING) {
```
rjwright: Will want to remove this.

src/dom-matrix-readonly.js
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    var a = matrix._matrix[0];
+    var b = matrix._matrix[1];
+    var c = matrix._matrix[2];
+    var d = matrix._matrix[3];
+    var e = matrix._matrix[4];
+    var f = matrix._matrix[5];
+    // See documentation http://www.w3.org/TR/css3-transforms/#MatrixDefined.
+    return new DOMMatrixReadonly([a, b, 0, 0,
```
rjwright: Move the first row onto a new line?

rjwright:
LGTM. I am fine with matricesApproxEqual where you have put it. It's very useful and very matrix-specific so it doesn't seem out of place there IMO.

src/dom-matrix-readonly.js
wilddamon: Done.

src/dom-matrix-readonly.js
wilddamon: Eh, I forgot that I did eventually add something to the testing scope lol. I created a new internal.testing scope for it, WDYT?

rjwright:
lgtm

PR #97, opened 2016-07-05
Add support for getting unimplemented types

wilddamon:
This add support for getting a base CSSStyleValue from the StylePropertyMaps. I wasn't sure how to allow construction of base CSSStyleValues for this purpose but disallow it for developers using this. Any suggestions?

Also fixed bug where an exception is thrown in Inline StylePropertyMap get method if the property is unset.

Note also that I have used "cssText" instead of "cssString" in this patch so that when #91 is merged it won't also need to be renamed.


rjwright:
Looks good. I see what you mean about exposing the CSSStyleValue constructor. What happens if you put it on internal?

wilddamon:
Hmm, well if we put it on internal, developers wouldn't be able to write 

``` javascript
if (someVariable instanceof CSSStyleValue)
```

Which I think we want to allow. I think this is a good compromise between similarity and what's actually possible - we could write 

``` javascript
function CSSStyleValue(cssText, yesIReallyMeanToCallThisIUnderstandItWontWorkForReals) {
  if (!cssText || !yesIReallyMeanToCallThisIUnderstandItWontWorkForReals) {
    throw new TypeError('Constructing CSSStyleValues will not be available in the native implementation');
  }
  this.cssText = cssText;
}
```

(well, with a more sensible variable name) to make it harder I guess?

src/css-style-value.js
```diff
@@ -70,5 +70,15 @@
...
   };
 
+
+  internal.CSSStyleValue = function(cssText) {
+    if (!cssText) {
+      throw new TypeError('CSSStyleValue must have a value');
+    }
+    this.cssText = cssText;
+  }
+  internal.inherit(internal.CSSStyleValue, CSSStyleValue);
```
rjwright: I think we should change this back to just setting the prototype, so that the constructor is fully hidden.

wilddamon:
I changed it to being in a separate file as discussed offline. I think this does what we want:

``` javascript
>> s = new typedOM.internal.CSSStyleValue('foo')
CSSStyleValue {cssText: "foo"}
>> s.__proto__
CSSStyleValue {}
>> s.__proto__.constructor
CSSStyleValue() {
    throw new TypeError('CSSStyleValue cannot be instantiated.');
  }
>> s.cssText
"foo"

>> new CSSStyleValue("foo")
css-style-value.js:18 Uncaught TypeError: CSSStyleValue cannot be instantiated.(…)

>> k = new CSSKeywordValue("bar")
CSSKeywordValue {keywordValue: "bar", cssText: "bar"}
{ cssText: "bar"
keywordValue: "bar"
__proto__: CSSStyleValue }
>> k.__proto__.__proto__.constructor
CSSStyleValue() {
    throw new TypeError('CSSStyleValue cannot be instantiated.');
}
```

rjwright:
LGTM! :D

PR #99, opened 2016-07-06
Remove the typedOMTesting scope

wilddamon:
This scope was set to window, which made it really hard to reason about, and made tests unrealistic. Since there wasn't actually anything useful being put on it, I've removed it altogether.


rjwright:
lgtm

PR #100, opened 2016-07-07
Implements a consuming parser for length types, moves length parsing to its own file

wilddamon:
This builds on https://github.com/css-typed-om/typed-om/pull/94 to some extent - probably don't review parsing.js until that one is closed.


shans:
lgtm

PR #101, opened 2016-07-08
Make is2D a property, not a method

wilddamon:
Along the way, move  _computeMatrix and _generateCssString methods to be really private.

### DIFF
--- a/src/css-rotation.js
+++ b/src/css-rotation.js
@@ -1,7 +1,7 @@
 // Copyright 2015 Google Inc. All rights reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
-// this.you may not use this file except in compliance with the License.
+// you may not use this file except in compliance with the License.
 //     You may obtain a copy of the License at
 //
 // http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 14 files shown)


src/css-rotation.js
```diff
@@ -76,17 +51,46 @@
...
+    }
+
+    for (var i = 0; i < arguments.length; i++) {
+      if (typeof arguments[i] != 'number') {
+        throw new TypeError('CSSRotation arguments must be of type \'number\'.');
+      }
+    }
+
+    this.is2D = (arguments.length == 1);
+    this.angle = this.is2D == 1 ? x : angle;
```
rjwright: Should just be `this.is2D ? x : angle;`

src/css-skew.js
```diff
@@ -14,6 +14,20 @@
 
 (function(internal, scope) {
 
+  function computeMatrix(cssSkew) {
+    // CSSSkew represented by the 2D matrix:
+    //   1     tan(ax)  0
+    // tan(ay)    1     0
+
+    var tanAx = CSSSkew._tanDegrees(cssSkew.ax);
```
rjwright, line 22: Shall we also take _tanDegrees out of CSSSkew?

rjwright:
LGTM with nits

src/css-rotation.js
wilddamon: Yes, done.

src/css-skew.js
wilddamon, line 22: That's actually a little more complicated than it looks due to the way that skew is tested. I'll do it in a separate PR.

PR #102, opened 2016-07-08
Add CSSTokenStreamValue

anthony-huang:


src/css-token-stream-value.js
```diff
@@ -0,0 +1,32 @@
...
+        //       https://drafts.css-houdini.org/css-typed-om/#tokenstreamvalue-normalization
+    }
+
+    internal.inherit(CSSTokenStreamValue, CSSStyleValue);
+
+    CSSTokenStreamValue.prototype.size = function() {
+        return this.values.length;
+    }
+
+    CSSTokenStreamValue.prototype.referenceAtIndex = function(index) {
```
wilddamon: Wasn't this method just an implementation detail in the C++ version? You can implement iteration in JS using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators

src/css-token-stream-value.js
```diff
@@ -0,0 +1,69 @@
...
+            throw new TypeError('CSSTokenStreamValue should be an array of string or CSSVariableReferenceValue');
+        }
+        for (var i = 0; i < values.length; i++) {
+            if (typeof values[i] != 'string' && !(values[i] instanceof CSSVariableReferenceValue)) {
+                throw new TypeError("CSSTokenStreamValue's elements should be string or CSSVariableReferenceValue");
+            }
+        }
+        this.listOfReferences = values;
+    }
+
```
wilddamon: Remove this newline

src/css-token-stream-value.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+        if (!Array.isArray(values)) {
+            throw new TypeError('CSSTokenStreamValue should be an array of string or CSSVariableReferenceValue');
+        }
+        for (var i = 0; i < values.length; i++) {
+            if (typeof values[i] != 'string' && !(values[i] instanceof CSSVariableReferenceValue)) {
+                throw new TypeError("CSSTokenStreamValue's elements should be string or CSSVariableReferenceValue");
+            }
+        }
+        this.listOfReferences = values;
```
wilddamon: This should be a "private" variable. In javascript, since you can't really have real private variables, we mark things as "private" by using an underscore at the start of the name:

``` javascript
this._listOfReferences = values;
```

src/css-token-stream-value.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    internal.inherit(CSSTokenStreamValue, CSSStyleValue);
+
+    CSSTokenStreamValue.prototype.size = function() {
```
wilddamon: This function is not required - it's another one that was an implementation detail of the C++ version :)

src/css-token-stream-value.js
```diff
@@ -0,0 +1,69 @@
...
+        return this.listOfReferences.length;
+    }
+
+    CSSTokenStreamValue.prototype.referenceAtIndex = function(index) {
```
wilddamon: This function is not required - it's another one that was an implementation detail of the C++ version :)

wilddamon:
Javascript style is to use 2 space indentation (you've used 4 spaces here). Could you please clean that up?

src/css-token-stream-value.js
```diff
@@ -0,0 +1,69 @@
...
+
+    CSSTokenStreamValue.prototype.keys = function() {
+        var nextIndex = 0;
+        var values = this.listOfReferences;
+
+        return {
+            next: function() {
+                return nextIndex < values.length ?
+                    { done: false, value: nextIndex++ } :
+                    { done: true, value: undefined };
```
wilddamon: This iterator code is very repetitive. Would it be worth making a helper function in src/util.js?

anthony-huang: I have changed this one to use the helper function, but I'm not sure whether using the eval() function is good or not. :cry:

src/css-token-stream-value.js
anthony-huang: done :smile:

src/css-token-stream-value.js
anthony-huang: done :smile:

src/util.js
```diff
@@ -57,8 +57,34 @@
...
+  /*
+   * Get iterator of object which returns a { done: false, value: expression }
+   *   if it has not reached the end of the object, or otherwise it returns
+   *   { done: true, value: undefined }.
+   * @param {!Object} object: The object to iterate over.
+   * @param {String} expression: The expression that will be executed if the
+   *   iterator has not reached the end of the array. expression contains the
+   *   format that we want to get as the value.
+   */
+
```
wilddamon: Remove this newline

src/util.js
```diff
@@ -57,8 +57,34 @@
...
+  function iterator(object, expression) {
+    if (!(this instanceof arguments.callee))
+      return new arguments.callee(object, expression);
+    var index = 0;
+
+    this.next = function next() {
+      if (index < object.length) {
+          var key = index;
+          var value = object[index++];
+          return { done: false, value: eval(expression) };
```
wilddamon: We shouldn't use eval() where possible - it will cause security warnings when people run the code. Perhaps you could try using objects as enums?

test/js/css-token-stream-value.js
```diff
@@ -0,0 +1,69 @@
+suite('CSSTokenStreamValue', function() {
+  test("CSSTokenStreamValue is a CSSTokenStreamValue and CSSStyleValue", function() {
+    assert.instanceOf(new CSSTokenStreamValue(), CSSTokenStreamValue,
+      'A new CSSTokenStreamValue should be an instance of CSSTokenStreamValue');
```
wilddamon: These messages are unnecessary - the default error message for assert.instanceOf is sufficient.

src/css-token-stream-value.js
wilddamon: Yeah, probably not... if you can think of another way that'd be great!

wilddamon:
This is starting to look really nice! Just need to figure out a way to no use eval :)

anthony-huang:
I suddenly thought of using a functor instead of eval(), so that we need to pass a function that will return the expected format. I don't know if this will be good enough.

The thing that must be considered is whether every keys(), values(), and entries() will only depend on key and value or not. If yes, then I think this will work :v

test/js/css-token-stream-value.js
```diff
@@ -0,0 +1,67 @@
+suite('CSSTokenStreamValue', function() {
+  test("CSSTokenStreamValue is a CSSTokenStreamValue and CSSStyleValue", function() {
+    assert.instanceOf(new CSSTokenStreamValue(), CSSTokenStreamValue);
+    assert.instanceOf(new CSSTokenStreamValue(), CSSStyleValue);
+  });
+
+  test('Constructor does not throw', function() {
```
wilddamon: It would be better to assert that the constructed object contains the correct values, instead of just testing that it doesn't throw

src/util.js
```diff
@@ -57,8 +57,32 @@
...
+      return new arguments.callee(object, callback);
+    var index = 0;
+
+    this.next = function next() {
+      if (index < object.length) {
+          var key = index;
+          var value = object[index++];
+          return { done: false, value: callback(key, value) };
+      } else return { done: true, value: undefined };
+    };
```
wilddamon: I think this can be done more clearly by just using an object:

``` javascript
function iterator(object, callback) {
  return {
    index: 0,
    next: function() {
      if (index < object.length) {
        return { done: false, value: callback(this.index, object[this.index++]) };
       }
       return { done: true, value: undefined };
    }
  }
}
```

src/util.js
```diff
@@ -57,8 +57,32 @@
     return false;
   }
 
+
+  /*
+   * Get iterator of object which returns a { done: false, value: expression }
+   *   if it has not reached the end of the object, or otherwise it returns
+   *   { done: true, value: undefined }.
+   * @param {!Object} object: The object to iterate over.
+   * @param {Function} callback: A function to be called by using key and value
```
wilddamon: callback: Defines the value attribute for the iterator. The function takes two arguments (key, value), and should return one of the following: key, value, or an array containing one or more of (key, value). Generally, key is used for the keys() method, value is used for the values() method, and [key, value] for the entries() method.

src/css-token-stream-value.js
```diff
@@ -0,0 +1,33 @@
...
+        throw new TypeError("CSSTokenStreamValue's elements should be string or CSSVariableReferenceValue");
+      }
+    }
+    this._listOfReferences = values;
+  }
+  internal.inherit(CSSTokenStreamValue, CSSStyleValue);
+
+  CSSTokenStreamValue.prototype.keys = function() {
+    return internal.objects.iterator(this._listOfReferences, function(key, value) { return key; });
+  }
```
wilddamon: Needs a semicolon; anything that is

``` javascript
something = function() {};
```

should have a semicolon at the end.

Functions defined like this:

``` javascript
function something() {
}
```

do not require a semicolon.

src/css-token-stream-value.js
```diff
@@ -0,0 +1,33 @@
...
+    return internal.objects.iterator(this._listOfReferences, function(key, value) { return value; });
+  }
+
+  CSSTokenStreamValue.prototype.entries = function() {
+    return internal.objects.iterator(this._listOfReferences, function(key, value) { return [key, value]; });
+  }
+
+  scope.CSSTokenStreamValue = CSSTokenStreamValue;
+
+}) (typedOM.internal, window);
```
wilddamon: No space between function and its call (`})(typedOM`)

src/util.js
wilddamon: Don't need a semicolon here (see earlier comment)

wilddamon:
LGTM. As discussed, next time please make a separate PR for unrelated changes!

PR #103, opened 2016-07-08
Remove the asMatrix method, replace it with matrix property that is a DOMMatrixReadonly.

wilddamon:
This PR is dependent on https://github.com/css-typed-om/typed-om/pull/101.


rjwright:
LGTM

PR #106, opened 2016-07-14
Implement parsing for CSSPositionValue

wilddamon:
This PR is dependent on both of
css-typed-om#103
css-typed-om#100

I put this up early because I was losing track of what I was working on... I'll assign it when it's ready for review.

This PR also contains these changes:
- Make cssText generator function in CSSPositionValue private
- Make separator optional for consumeRepeated

### DIFF
--- a/src/css-position-value-parsing.js
+++ b/src/css-position-value-parsing.js
@@ -22,7 +22,6 @@
     if (!params || params[0].length != 2) {
       return;
     }
-    console.log(params);
     var lengths = params[0];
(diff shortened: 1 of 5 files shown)


rjwright:
LGTM.

As discussed offline, we will make _parsePositionValue_ and the _from_ method in other CSSStyleValue subtypes (e.g. _CSSColorValue, CSSNumberValue_) consistent with each other in a future change, and wire _parsePositionValue_ up to _CSSStyleValue.parse_.

PR #107, opened 2016-07-14
Alphabetise target-config as much as possible.

wilddamon:


wilddamon:
This one should be easy :)

shans:
lgtm

PR #108, opened 2016-07-14
Add a consumer for CSSAngleValues.

wilddamon:
Also makes the output of consumeList consistent with the other consumers.


shans:
lgtm

PR #111, opened 2016-07-18
Add CSSResourceValue

anthony-huang:


src/css-resource-value.js
```diff
@@ -0,0 +1,26 @@
+(function(internal, scope) {
+
+  function CSSResourceValue(state) {
```
wilddamon: Put this after `CSSResourceState` and `isValidResourceState`.

wilddamon:
LGTM with nits. Will merge after you fix :)

src/css-resource-value.js
anthony-huang: done :)

PR #112, opened 2016-07-18
Add parsing for CSSMatrix

wilddamon:
Added a `parsing.optional` method which can be used with consumeList to bring in optional arguments (in this case the "3d" part of matrix/matrix3d).

### DIFF
--- a/src/css-matrix-parsing.js
+++ b/src/css-matrix-parsing.js
@@ -1,0 +1,43 @@
+// Copyright 2016 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+//     You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


src/css-matrix-parsing.js
```diff
@@ -0,0 +1,43 @@
...
+        parsing.optional(parsing.consumeToken.bind(null, /^3d/i)),
+        parsing.ignore(parsing.consumeToken.bind(null, /^\(/)),
+        parsing.consumeRepeated.bind(null, parsing.consumeNumber, /^,/),
+        parsing.ignore(parsing.consumeToken.bind(null, /^\)/)),
+    ], string);
+    if (!params) {
+      return null;
+    }
+    var leftover = params[1];
+    params = params[0];
```
rjwright: Kinda confusing. Can we call the consumeList output "result" and make `var params = result[0]`?

rjwright:
LGTM w/ nit

src/css-matrix-parsing.js
wilddamon: Done.

PR #114, opened 2016-07-21
Add parsing for CSSPerspective

wilddamon:
Also expose consumeSimpleLength, and make sure consumeSimpleLength's unit parsing is not case-sensitive.


shans:
LGTM

PR #119, opened 2016-07-26
Implement iteration for StylePropertyMapReadonly

wilddamon:
I needed to update the helper functions Anthony wrote to work with the other iteration methods available ([...spread] and for (let...of)).
Added tests for everything.


wilddamon:
Argh, right now tests are failing because I moved the number parsing in my other PR. Testing this might depend on getting parsing hooked up :frowning:

wilddamon:
This one is ready to look at now.

Random question - I made a test-util.js file in the testing directory, should it have a license header?

src/css-token-stream-value.js
```diff
@@ -16,16 +16,32 @@
   }
   internal.inherit(CSSTokenStreamValue, CSSStyleValue);
 
-  CSSTokenStreamValue.prototype.keys = function() {
-    return internal.objects.iterator(this._listOfReferences, function(key, value) { return key; });
+  CSSTokenStreamValue.prototype[Symbol.iterator] = function() {
```
samthor: what's the basic support level for the polyfill? I notice that `Symbol` isn't used anywhere else, and it's from ES6.

src/util.js
```diff
@@ -57,32 +59,57 @@
...
    *   { done: true, value: undefined }.
-   * @param {!Object} object: The object to iterate over.
-   * @param {Function} callback: Defines the value attribute for the iterator.
-   *   The function takes two arguments (key, value), and should return one of
-   *   the following: key, value, or an array containing one or more of
-   *   (key, value). Generally, key is used for the keys() method, value is
-   *   used for the values() method, and [key, value] for the entries() method.
+   * See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols
+   * for more information about iterators.
+   * @param {!Array} keys: An array of keys to iterate over.
```
samthor, line 70: `{!Array<string>}` ?

src/util.js
```diff
@@ -57,32 +59,57 @@
...
+   * @param {Function} callback: A function that returns the value for the
```
samthor, line 71: Rather than writing such a long description, you could instead say- `{function(string): (string|...)`

src/util.js
```diff
@@ -57,32 +59,57 @@
...
+          var key = keys[index++];
+          return { done: false, value: callback(key) };
+        } else {
+          return { done: true, value: undefined };
+        }
       }
-    }
+    };
+    result[Symbol.iterator] = function() { return result; };
+    return result;
```
samthor, line 92: small nit: you could hide `next` rather than exposing it, if `Symbol.iterator` is the intended use

src/util.js
samthor, line 70: Oh, it's not necessarily a string. Whoops, I thought this was specifically about typed OM things, but I see this is a util file.

src/util.js
```diff
@@ -57,32 +59,57 @@
...
+   * @param {number} length: The length of the array object
+   * @param {Function} callback: A function that returns the value for the
+   *   iterator. The function takes a single argument: the current index, and
+   *   should return one of the following: index, value, or an array containing
+   *   one or more of (index, value). Generally, index is used for the keys()
+   *   method, value is used for the values() method, and [index, value] for
+   *   the entries() method.
+   */
+  function arrayIterator(length, callback) {
+    return iterator([...Array(length).keys()], callback);
```
samthor, line 107: using ... also implies ES6

src/style-property-map-readonly.js
```diff
@@ -34,7 +34,8 @@
     }
 
     if (internal.propertyDictionary().isSupportedProperty(property)) {
-      return CSSStyleValue.parse(property, propertyString);
+      var result = CSSStyleValue.parse(property, propertyString);
+      return result instanceof Array ? result : [result];
```
samthor: `Array.isArray(result)`? (src/style-property-map uses `instanceof Array`, src/css-...-value.js uses `Array.isArray`).

src/css-token-stream-value.js
wilddamon: ¯_(ツ)_/¯ Iteration itself is ES6, and it specced in Typed OM. I suspect that the problem where "CSS" can't be found is probably the same issue...

samthor:
LGTM

...with the note that this breaks support for ES5 browsers like IE11 and such (I guess that may be the whole point, sorry I missed that initially).

src/style-property-map-readonly.js
wilddamon: Done

src/css-token-stream-value.js
wilddamon: I could make the declaration of iterator support gated on the existence of Symbol though? At least then they could use the rest of the polyfill...

src/util.js
wilddamon: Done.

src/util.js
wilddamon: ¯\_(ツ)_/¯ iteration is ES6 anyway

src/css-token-stream-value.js
wilddamon: Actually, I changed my mind. I'll just add an item to our task list to do support for non-ES6 browsers, and we can do transpiling later.

src/util.js
wilddamon: I'm not sure that would be useful? It'd be function(object):object|array, which isn't especially descriptive.

PR #120, opened 2016-07-27
Remove the .from() parsing methods that don't exist in the spec

wilddamon:
Sadly I had to disable several tests until the parsing is hooked up again (properly this time).

Also:
- Moved parsing functionality into -parsing.js files to be consistent with other parsing code.
- Got rid of the isInt function and replaced with Number.isInteger method.


shans:
lgtm

PR #121, opened 2016-07-27
Update isValidLengthType CSSLengthValue types to be internal

wilddamon:
This leaves the API more similar to what is specced for Lengths.


src/css-length-value.js
```diff
@@ -123,5 +122,8 @@
     throw new TypeError('Should not be reached');
   };
 
+  internal.CSSLengthTypes = CSSLengthTypes;
+  internal.isValidLengthType = isValidLengthType;
+
   scope.CSSLengthValue = CSSLengthValue;
```
rjwright, line 128: This is not introduced in your change, but I just noticed it. The spec doesn't expose a CSSLengthValue constructor, only `from`:

https://drafts.css-houdini.org/css-typed-om-1/#lengthvalue-objects

rjwright:
lgtm

PR #124, opened 2016-07-27
Move individual class parsing files into a subfolder

wilddamon:
Note that this will probably conflict with other PRs.

I thought the number of files was getting a bit unwieldy, so I put the parsing files into a subfolder. WDYT?


rjwright:
Did I mention this to you, or do great minds simply think alike?

wilddamon:
You didn't mention it to me - I guess great minds really do think alike!

rjwright:
LGTM

PR #126, opened 2016-07-28
Make CSSSkew work with CSSAngleValue arguments

rjwright:
This spec change https://github.com/w3c/css-houdini-drafts/pull/271 means we won't need to store _ax/_ay and ax/ay.

Will update cssText in other TransformComponenets in a later patch.

### DIFF
--- a/src/css-skew-parsing.js
+++ b/src/css-skew-parsing.js
@@ -23,11 +23,11 @@
     switch (type) {
       case 'x':
         var result = [new CSSSkew(angles[0], zeroAngle), remaining];
-        result[0].cssText = 'skewx(' + result[0]._ax.cssText + ')';
+        result[0]._inputType = 'x';
         return result;
       case 'y':
         var result = [new CSSSkew(zeroAngle, angles[0]), remaining];
-        result[0].cssText = 'skewy(' + result[0]._ay.cssText + ')';
+        result[0]._inputType = 'y';
         return result;
     }
   }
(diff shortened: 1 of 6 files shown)


rjwright:
I should probably also add some CSSAngleValue.cssText tests in this change.

src/css-skew-parsing.js
```diff
@@ -19,11 +19,16 @@
     if (angles.length != 1) {
       return null;
     }
+    var zeroAngle = new CSSAngleValue(0, 'deg');
     switch (type) {
       case 'x':
-        return [new CSSSkew(angles[0].degrees, 0), remaining];
+        var result = [new CSSSkew(angles[0], zeroAngle), remaining];
+        result[0].cssText = 'skewx(' + result[0]._ax.cssText + ')';
```
wilddamon: We probably shouldn't be looking at private members or setting cssText (which will be a readonly member when implemented in Blink). It's setting a bad example for how you can use Typed OM (even though these are internal methods). Can we instead make a method on typedOM.internal that sets the cssText of a CSSStyleValue? (even if the method does the same thing I think it looks less weird).

Also in these cases, the cssText will be the consumed string right? I think it would be better to look at how many chars were consumed, and fetch the original parsed string, and use that (possibly lowercased), rather than constructing it again here.

rjwright: > We probably shouldn't be looking at private members or setting cssText (which will be a readonly member when implemented in Blink). It's setting a bad example for how you can use Typed OM (even though these are internal methods). Can we instead make a method on typedOM.internal that sets the cssText of a CSSStyleValue? (even if the method does the same thing I think it looks less weird).

Yes, good idea.

> Also in these cases, the cssText will be the consumed string right? I think it would be better to look at how many chars were consumed, and fetch the original parsed string, and use that (possibly lowercased), rather than constructing it again here.

Yeah. I do want to normalize it, and I needed to get rid of the trailing chars. But you are right that getting rid of trailing chars isn't hard. Lower casing will partly normalize it, but spaces after separators won't be normalized. Do we care?

PR #129, opened 2016-07-29
Add CSSAngleValue support to CSSSkew constructor

rjwright:
Make CSSSkew work with CSSAngleValue arguments. Slight improvements to tests.

Test case `cssText: 'skew(30.0000002521989deg, 179.99999979432deg)'` will be improved in a future CL which will change cssText to reflect the input angles.


wilddamon:
The Rotation stuff probably shouldn't be in this PR. LGTM once it's removed (looks like a simple revert)

PR #130, opened 2016-07-29
Make cssText for CSSRotation and CSSSkew reflect the input arguments, and related tweaks

rjwright:
Make cssText for CSSRotation and CSSSkew reflect the input arguments (or input string, in the case of the parser). Also make the getter lazy, and add tests for cssText.

See the (spec section)[https://drafts.css-houdini.org/css-typed-om-1/#csstransformvalue] on CSSTransformComponent cssText.

### DIFF
--- a/src/css-rotation-parsing.js
+++ b/src/css-rotation-parsing.js
@@ -19,17 +19,25 @@
     if (numbers.length != 4) {
       return null;
     }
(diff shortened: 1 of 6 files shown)


rjwright:
Future PRs will make other CSSTransformComponents match this behavior.

wilddamon:
LGTM

PR #131, opened 2016-08-01
Fix problems with CSSRotation.cssText and delete duplicate test.

rjwright:
Not sure how these mistakes got in there to begin with. Sorry.


wilddamon:
lgtm

PR #132, opened 2016-08-02
[WIP] Implement part of hooking up all the new parsing code

wilddamon:
This definitely shouldn't be merged as-is...

I found that testing my other PRs depended on this, so I made a start. This doesn't quiiiite work yet, want to finish it for me? :stuck_out_tongue:


src/parsing/css-style-value-parsing.js
```diff
@@ -0,0 +1,74 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(internal) {
+
+  var parsers = {};
+  parsers[CSSNumberValue] = internal.parsing.consumeNumber;
```
wilddamon, line 18: It'd probably be better to use a type enum or use a proper map (https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Map) here instead. It stringifies the function if you do this:

``` javascript
>> var o = {}
>> o[CSSStyleValue] = 5
>> Object.keys(o)
["function CSSStyleValue() {\n  throw new TypeError('CSSStyleValue cannot be instantiated.');\n}"]
```

src/property-dictionary.js
```diff
@@ -50,14 +50,14 @@
     return (this._listSeparator.hasOwnProperty(property));
   };
 
-  PropertyDictionary.prototype.getValidStyleValuesArray = function(property) {
+  PropertyDictionary.prototype.supportedStyleValues = function(property) {
```
wilddamon, line 53: Trying to be less verbose here. I didn't update the tests for this naming change yet.

src/parsing/css-style-value-parsing.js
```diff
@@ -0,0 +1,74 @@
...
+  parsers[CSSLengthValue] = internal.parsing.consumeLength;
+  parsers[CSSTransformValue] = internal.parsing.consumeTransform;
+
+  function consume(consumer, string) {
```
wilddamon, line 22: This is a helper function to deal with the fact that there may not be an entry in the parsers map above, and avoid writing if (!function) everywhere. Otherwise below you have to write (which isn't really that bad I suppose). Anyway something to think about.

``` javascript
for (var i = 0; i < supportedStyleValues.length; i++) {
  var consumer = parsers[supportedStyleValues[i]];
  var parsed = null;
  if (consumer) {
    parsed = consumer(property, string);
  }
  if (parsed) {
    return parsed;
  }
}
```

If we keep it, should probably give this a more descriptive name, could be "maybeConsume"?

There's a bug in this function anyway, it's not passing in the property.

src/parsing/css-style-value-parsing.js
```diff
@@ -0,0 +1,74 @@
...
+    if (!consumer) {
+      return null;
+    }
+    return consumer(string);
+  }
+
+  function consumeKeyword(property, string) {
+    var consumed = internal.parsing.consumeToken(/^[a-z_]+[a-z0-9_-]*/i, string);
```
wilddamon, line 30: This regex is meant to be a representation of https://www.w3.org/TR/css-syntax-3/#ident-token-diagram

/^ anchor start of string
[a-z_]+ one or more of alphabetical characters or underscore
[a-z0-9_-]\* zero or more of alphanumeric, underscore, or dash
/i case insensitive for alphabetical characters

src/parsing/css-style-value-parsing.js
```diff
@@ -0,0 +1,74 @@
...
+    return [new CSSKeywordValue(consumed[0]), consumed[1]];
+  }
+
+  function consumeStyleValue(property, string) {
+    if (!internal.propertyDictionary().isSupportedProperty(property)) {
+      // TODO: How do custom properties play into this?
+      return null;
+    }
+
+    var consumedKeyword = consumeKeyword(property, string);
```
wilddamon, line 46: I'm assuming keywords are more common than StyleValues here. I don't really know how to determine the actual frequency of keyword vs StyleValue, but if we decided that it was more likely to be a StyleValue, we can change the order.

src/property-dictionary.js
```diff
@@ -78,8 +78,7 @@
     return (lengthValue.type == 'percent');
   };
 
-  PropertyDictionary.prototype.
-      isValidKeyword = function(property, styleValueString) {
+  PropertyDictionary.prototype.isValidKeyword = function(property, styleValueString) {
```
wilddamon, line 81: Removed newline: I find this style more readable.

Also in this function, we should always return true for globally valid keywords (e.g. initial, inherit, etc)

src/parsing/css-style-value-parsing.js
rjwright, line 22: > This is a helper function to deal with the fact that there may not be an entry in the parsers map above

Since it's only called once I might remove it.

> There's a bug in this function anyway, it's not passing in the property.

Why would it need to pass in the property?

wilddamon, line 22: My mistake, the problem was consumeLength vs consumeLengthValue, which I see you already fixed in your branch :)

Issue #78, opened 2016-05-01
Testing instructions in readme don't seem to work


rjwright:
$ npm test

> typed-om@1.0.0 test /home/rjwright/TOM/typed-om
> ./node_modules/phantomjs/bin/phantomjs ./node_modules/mocha-phantomjs-core/mocha-phantomjs-core.js ./test/runner.html

sh: ./node_modules/phantomjs/bin/phantomjs: No such file or directory
npm ERR! weird error 127
npm WARN This failure might be due to the use of legacy binary "node"
npm WARN For further explanations, please read
/usr/share/doc/nodejs/README.Debian

npm ERR! not ok code 0

wilddamon:
I referenced the wrong bug in my PR. Reopening.

samthor:
The issue here partially is because you haven't run `npm install` to get PhantomJS. I'll update the README.

samthor:
Oh, the README includes it already. @rjwright your error looks like it just wasn't installed. If this isn't the case then please reopen.

rjwright:
I just switched to using the test runner.  If it's working for you then we can leave it closed for now.

PR #136, opened 2016-08-10
Make including typed om work when running locally and when using http

wilddamon:
Fixes https://github.com/css-typed-om/typed-om/issues/78

I tested this using the following directory structure and code:

```
outerdir/typed-om
outerdir/test.html
```

``` html
<script src='typed-om/typed-om.js'></script>
<script>
document.onreadystatechange = function () {
    if (document.readyState == "complete") {
      var matrix = new window.DOMMatrixReadonly([5, 5, 5, 5, 5, 5]);
      console.log(matrix);
    }
};
</script>
```

Then I opened outerdir/test.html using file:///path/to/outerdir/test.html, and also by running a python http server in outerdir:

``` python
python -m SimpleHTTPServer 6060
```

then opening localhost:6060/test.html.

### DIFF
--- a/typed-om.js
+++ b/typed-om.js
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 var TYPED_OM_TESTING = false;
-var typedOMIncludePath = typedOMIncludePath || '/typed-om';
+var typedOMIncludePath = typedOMIncludePath || 'typed-om';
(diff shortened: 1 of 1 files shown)


rjwright:
lgtm

PR #137, opened 2016-08-16
Wire up consumers & make all consumers return null on failure

rjwright:
WIP because it needs more tests.

This is an evolution of Eddy's work here https://github.com/css-typed-om/typed-om/pull/132.

We had some consumers returning null and some returning undefined on failure. I've changed them so that they all return null (https://drafts.css-houdini.org/css-typed-om-1/#dom-cssstylevalue-parse).

### DIFF
--- a/src/parsing/css-style-value-parsing.js
+++ b/src/parsing/css-style-value-parsing.js
@@ -19,13 +19,6 @@
   parsers['CSSLengthValue'] = internal.parsing.consumeLengthValue;
   parsers['CSSTransformValue'] = internal.parsing.consumeTransformValue;
 
-  function maybeConsume(consumer, string) {
-    if (!consumer) {
-      return null;
-    }
-    return consumer(string);
-  }
-
   function consumeKeyword(property, string) {
     var consumed = internal.parsing.consumeToken(/^[a-z_]+[a-z0-9_-]*/i, string);
     if (!consumed) {
@@ -50,8 +43,8 @@
 
     var supportedStyleValues = internal.propertyDictionary().supportedStyleValues(property);
     for (var i = 0; i < supportedStyleValues.length; i++) {
-      var parsed = maybeConsume(parsers[supportedStyleValues[i].name], string);
-      if (parsed) {
+      var parsed, consumer = parsers[supportedStyleValues[i].name];
+      if (consumer && (parsed = consumer(string))) {
         return parsed;
       }
     }
--- a/src/parsing/css-style-value-parsing.js
+++ b/src/parsing/css-style-value-parsing.js
@@ -1,0 +1,74 @@
+// Copyright 2016 Google Inc. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+//     You may obtain a copy of the License at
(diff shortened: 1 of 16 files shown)


src/parsing/css-style-value-parsing.js
```diff
@@ -0,0 +1,66 @@
...
+
+    var consumedKeyword = consumeKeyword(property, string);
+    if (consumedKeyword) {
+      return consumedKeyword;
+    }
+
+    var supportedStyleValues = internal.propertyDictionary().supportedStyleValues(property);
+    for (var i = 0; i < supportedStyleValues.length; i++) {
+      var parsed, consumer = parsers[supportedStyleValues[i].name];
+      if (consumer && (parsed = consumer(string))) {
```
wilddamon: It's a bit confusing to have the assignment inside the if-statement.  I think it's clearer to do

``` javascript
for (var i = 0; i < supportedStyleValues.length; i++) {
  var consumer = parsers[supportedStyleValues[i].name];
  if (!consumer) {
    return null;
  }
  var parsed = consumer(string);
  if (parsed) {
    return parsed;
  }
}
```

even if it's longer.

src/parsing/parsing.js
```diff
@@ -128,7 +130,10 @@
       }
     }
     var parsed = parser(string.substr(0, n));
-    return parsed == undefined ? undefined : [parsed, string.substr(n)];
```
wilddamon, line 131: lol what, did I write that?

If you want to keep the ternary and still make it less bad, it could be

``` javascript
return parsed ? [parsed, string.substr(n)] : null;
```

which is much easier to understand.

wilddamon:
This is looking great, looking forward to seeing the tests!

src/parsing/css-style-value-parsing.js
rjwright: I thought you might say that :P It will actually need to be

```
for (var i = 0; i < supportedStyleValues.length; i++) {
  var consumer = parsers[supportedStyleValues[i].name];
  if (!consumer) {
    continue;
  }
  var parsed = consumer(string);
  if (parsed) {
    return parsed;
  }
}
```

or

```
for (var i = 0; i < supportedStyleValues.length; i++) {
  var consumer = parsers[supportedStyleValues[i].name];
  if (consumer) {
    var parsed = consumer(string);
    if (parsed) {
        return parsed;
    }
  }
}
```

src/parsing/parsing.js
rjwright, line 131: I thought about it, but other places seem to do what I have done rather than use a ternary.

rjwright, line 131: Still other places do

if (!parsed) {return null;}
return ...;

src/parsing/css-style-value-parsing.js
wilddamon: I think the first one is the most readable, but both seem fine to me :)

src/parsing/parsing.js
wilddamon, line 131: Oh. Well either way should be fine then. I wouldn't bother going and making them consistent.

wilddamon:
Nice work! LGTM

PR #138, opened 2016-08-22
Add CSSImageValue object and tests

anthony-huang:


src/css-image-value.js
```diff
@@ -0,0 +1,39 @@
+(function(internal, scope) {
+
+  function CSSImageValue(image) {
+    if (!(image instanceof Image)) {
```
wilddamon, line 13: CSSImageValue should take a string URL: https://drafts.css-houdini.org/css-typed-om/#imagevalue-objects

src/css-image-value.js
```diff
@@ -0,0 +1,39 @@
...
+      throw new TypeError("image must be an Image object");
+    }
+    this._image = image;
+
+    var imagesEventsHandler = {
+      onload: function() {
```
wilddamon: There's no need to wrap these functions inside another object. You can just write

``` javascript
function CSSImageValue(image) {
  ...
  function onLoad() {
  ...
  }
  ...
  this._image.onload = onLoad.bind(this);
```

src/css-image-value.js
anthony-huang: Hmm.. I think that is for the CSSURLImageValue. Or should I just make this as a CSSURLImageValue? Actually I want to make something like this: https://github.com/css-typed-om/typed-om/compare/master...anthony-huang:try_CSSImageValue

wilddamon: Oh right. The problem with this implementation is that users of the Polyfill will be able to access and construct CSSImageValues using this constructor. You'll need to hide the internal constructor using some code construct like this: https://github.com/css-typed-om/typed-om/pull/135/files

anthony-huang: Do I need to make the from() function? Or just make the one like https://github.com/css-typed-om/typed-om/pull/135/files#diff-0c45aef1150ce960e1a54ed56eddadcbR132 (from line 132 to 142)?

wilddamon: Since CSSImageValue does not have a from method, you don't need to do it, no. Your CSSImageValue here also shouldn't have a from method. It should only expose methods that are in the spec.

test/js/css-image-value.js
```diff
@@ -0,0 +1,28 @@
+suite('CSSImageValue', function() {
```
wilddamon, line 1: You should test somewhere that a developer can query an instance of ImageValue using the "instanceof" operator, and get the right result with the exposed CSSImageValue constructor, e.g.

``` javascript
if (a instanceof CSSImageValue)
```

or

``` javascript
if (a instanceof CSSResourceValue)
```

or

``` javascript
if (a instanceof CSSStyleValue)
```

There are various other tests which do similar things using the assert.instanceOf helper.

test/js/css-image-value.js
```diff
@@ -0,0 +1,34 @@
...
+  test('CSSImageValue object is a CSSImageValue, CSSResourceValue, and CSSStyleValue', function() {
+    assert.instanceOf(new typedOM.internal.CSSImageValue(new Image()), typedOM.internal.CSSImageValue);
```
wilddamon: Note that if developers are going to check something using instanceof, they don't have access to the internal classes. You may need to rearrange your function hiding thing to be the other way around.

anthony-huang: I'm still not sure. So should I still make the constructor that accepts an Image internal?

wilddamon: No, it's not related to the constructor argument. It's that this test should be

``` javascript
assert.instanceOf(new typedOM.internal.CSSImageValue(new Image()), CSSImageValue);
```

so that it's more consistent with what a developer would see when they do an instanceof check.

This may mean you need to change the order you define the constructors in.

test/js/css-image-value.js
anthony-huang: Done :)

wilddamon:
LGTM

PR #139, opened 2016-08-23
Remove support for double angles in CSSSkew and CSSRotation constructors

rjwright:
As per spec change here https://github.com/w3c/css-houdini-drafts/pull/280


wilddamon:
\o/ LGTM

PR #141, opened 2016-08-23
Update CSSTranslation.cssText to new style

rjwright:
Make CSSTranslation.cssText reflect input text for parsed values & make it lazy.

Makes it consistent with https://github.com/css-typed-om/typed-om/pull/130/files (rotate and skew).

Other types will follow in separate PRs.

### DIFF
--- a/src/css-translation.js
+++ b/src/css-translation.js
@@ -32,7 +32,9 @@
 
   function generateCssString(cssTranslation) {
     if (cssTranslation.is2D) {
-      return 'translate(' + cssTranslation.x.cssText + ', ' + cssTranslation.y.cssText + ')';
+      return 'translate('
+        + cssTranslation.x.cssText + ', '
+        + cssTranslation.y.cssText + ')';
     } else {
       return 'translate3d('
         + cssTranslation.x.cssText + ', '
@@ -71,6 +73,7 @@
       set: function(newCssText) {}
     });
   }
+  internal.inherit(CSSTranslation, internal.CSSTransformComponent);
 
   // These functions (cssTranslationFromTranslate*) are for making CSSTranslations from parsed CSS
   // Strings. These are needed for setting the cssText.
@@ -125,7 +128,6 @@
     return result;
   }
 
-  internal.inherit(CSSTranslation, internal.CSSTransformComponent);
   internal.cssTranslationFromTranslate = cssTranslationFromTranslate;
   internal.cssTranslationFromTranslate3d = cssTranslationFromTranslate3d;
   internal.cssTranslationFromTranslateX = cssTranslationFromTranslateX;
--- a/src/css-translation.js
+++ b/src/css-translation.js
@@ -31,15 +31,23 @@
   };
 
   function generateCssString(cssTranslation) {
-    var cssText;
-    if (cssTranslation.is2D) {
-      cssText = 'translate(' + cssTranslation.x.cssText + ', '
-          + cssTranslation.y.cssText + ')';
-    } else {
-      cssText = 'translate3d(' + cssTranslation.x.cssText + ', '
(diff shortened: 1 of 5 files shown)


src/parsing/css-translation-parsing.js
```diff
@@ -19,7 +19,9 @@
       if (coords.length != 3) {
         return null;
       }
-      return [new CSSTranslation(coords[0], coords[1], coords[2]), remaining];
+      var result = [new CSSTranslation(coords[0], coords[1], coords[2]), remaining];
+      result[0]._inputType = '3';
```
wilddamon: I really don't like this thing where we're modifying the private attributes of the result outside the css-translation.js file - this can get hard to reason about very quickly. What if instead we had internal-only functions `internal.buildCSSTranslation(x, y, z)` and `internal.buildCSSTranslationX(x)` etc, in css-translation.js, that attach the correct cssText method to the resulting object? I'm imagining something like this

``` javascript

function CSSTranslation(x, y, z) {...}
...
function buildCSSTranslationX(x) {
  var result = new CSSTranslation(x, zeroLength);
  Object.defineProperty(result, 'cssText', {
    get: function() { return cssTextX(this); },
    set: function(newText) {}
  });
  return result;
}

internal.buildCSSTranslationX = buildCSSTranslationX;
```

This also gets rid of the magic '1', '2', 'x' etc, which I also think is not so great.

rjwright: I think that might be more confusing, and add a lot of code. What about if we just set _cssText directly from the parser? That gets rid of the magic values.

rjwright: > I think that might be more confusing, and add a lot of code. What about if we just set _cssText directly from the parser? That gets rid of the magic values.

Actually I can't completely get rid of the magic flags, but I won't have to store them. And I can put them in an enum.

rjwright:
@wilddamon after talking to @shans and deciding we don't want to normalize cssText for parsed values, I have simplified the code and gotten rid of the magic flags.

I tried adding a factory for each input type, but I think was less readable. I also don't think we should move this logic into the class file, because it is a special case for parsing.

rjwright:
> I tried adding a factory for each input type, but I think was less readable. I also don't think we should move this logic into the class file, because it is a special case for parsing.

But then I did it anyway, cause I don't really mind :P

src/css-translation.js
```diff
@@ -61,10 +62,78 @@
...
+        return this._cssText;
+      },
+      set: function(newCssText) {}
+    });
   }
   internal.inherit(CSSTranslation, internal.CSSTransformComponent);
 
+  // These functions (cssTranslationFromTranslate*) are for making CSSTranslations from parsed CSS
+  // Strings. These are needed for setting the cssText.
+  function cssTranslationFromTranslate(coords, string, remaining) {
```
wilddamon, line 80: This wasn't what I was thinking either - now you've moved parsing logic into the main file. You can definitely keep them separate - I'll send you an example in a bit.

src/parsing/css-translation-parsing.js
```diff
@@ -62,21 +39,17 @@
 
     switch (type) {
       case '3d' :
-        return translate3d(coords, remaining);
+        return internal.cssTranslationFromTranslate3d(coords, string, remaining);
```
wilddamon, line 42: This is incorrect anyway - this will put the entire input into the cssText of the CSSTranslation, instead of just the consumed part.

rjwright: > This is incorrect anyway - this will put the entire input into the cssText of the CSSTranslation, instead of just the consumed part.

That is true. Forgot about that. I have a feeling we went through all of this when we did rotate and skew.

PR #143, opened 2016-08-30
Add CSSURLImageValue object

anthony-huang:


test/js/css-url-image-value.js
```diff
@@ -0,0 +1,47 @@
...
+      assert.strictEqual(urlImageValue.state, "loaded");
+      assert.strictEqual(urlImageValue.intrinsicWidth, 1);
+      assert.strictEqual(urlImageValue.intrinsicHeight, 1);
+      assert.strictEqual(urlImageValue.intrinsicRatio, 1);
+      done();
+    }
+  });
+
+  test('Invalid image will have error state and null intrinsic dimensions', function(done) {
+    var urlImageValue = new CSSURLImageValue('http://localhost');
```
wilddamon: Just use something you know definitely won't return an image. (e.g. resources/nonexisting.png?) Who knows what localhost could return on someone else's machine!

test/js/css-url-image-value.js
```diff
@@ -0,0 +1,47 @@
+suite('CSSURLImageValue', function() {
```
wilddamon, line 1: The value of cssText isn't being tested anywhere.

anthony-huang: Done.

test/js/css-url-image-value.js
anthony-huang: Done.

wilddamon:
LGTM

PR #144, opened 2016-09-01
Enable set and get CSSURLImageValue from StyleMap for Background Image

anthony-huang:


src/parsing/css-url-image-value-parsing.js
```diff
@@ -0,0 +1,25 @@
...
+// limitations under the License.
+
+(function(internal) {
+
+  function consumeURLImageValue(string) {
+    // need to remove 'url("' and '")'
+    var temp = string.substring(5);
+    var clean = temp.substr(0, temp.length - 2);
+    return [[new CSSURLImageValue(clean)]];
+  }
```
wilddamon, line 29: This doesn't look right.

consume\* functions should return [ consumedThing, remainingString ]

src/parsing/css-url-image-value-parsing.js
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//     See the License for the specific language governing permissions and
+// limitations under the License.
+
+(function(internal) {
+
+  function consumeURLImageValue(string) {
+    // need to remove 'url("' and '")'
+    var temp = string.substring(5);
```
wilddamon: You can use the other parsing consume\* methods to clarify this and also fail correctly if the string isn't in the form url(stuff).

Take a look at the other parsers and see how they use the consume\* libraries, for example the handling for "matrix(1,2,3,4,5,6)".
 https://github.com/css-typed-om/typed-om/blob/master/src/parsing/css-matrix-parsing.js

Also, "temp" and "clean" are generally not good variable names - try to use more descriptive names where possible.

test/js/css-url-image-value.js
```diff
@@ -1,4 +1,13 @@
 suite('CSSURLImageValue', function() {
+  setup(function() {
+    this.element = document.createElement('div');
```
wilddamon: Instead of creating a new element, can you just use `document.documentElement` instead?

test/js/css-url-image-value.js
```diff
@@ -46,4 +55,23 @@ suite('CSSURLImageValue', function() {
...
+    var inlineStyleMap = this.element.styleMap();
+    inlineStyleMap.set("background-image", new CSSURLImageValue('resources/1x1-green.png'));
+    var urlImageValue = inlineStyleMap.get("background-image");
+    assert.instanceOf(urlImageValue, CSSURLImageValue);
+    assert.equal(urlImageValue.url, "resources/1x1-green.png");
+
+    var oldOnload = urlImageValue._image.onload;
+
+    urlImageValue._image.onload = function() {
+      oldOnload();
```
wilddamon: This is getting kind of repetitive. Can you create a function for chaining functions together? Something like this perhaps (maybe you can think of something more general?):

``` javascript
function chain(firstFn, secondFn, opt_this) {
  return function() {
    firstFn.call(opt_this);
    secondFn.call(opt_this);
  };
}
```

test/js/css-url-image-value.js
anthony-huang: Hmm.. If I use `document.documentElement`, the background of `runner.html` will also change. I can also use the `border-image-source` instead of `background-image`.

test/js/css-url-image-value.js
anthony-huang: I made a `function chain(listOfFunctions, opt_this)`, it will execute each function in listOfFunctions with opt_this

src/parsing/css-url-image-value-parsing.js
```diff
@@ -0,0 +1,32 @@
...
+
+    var result = parsing.consumeList([
+      parsing.ignore(parsing.consumeToken.bind(null, /^url\("/i)),
+      parsing.consumeToken.bind(null, /[^"]*/),
+      parsing.ignore(parsing.consumeToken.bind(null, /^"\)/))
+    ], string);
+    if (!result) {
+      return null;
+    }
+    return [[new CSSURLImageValue(result[0][0])], result[1]];
```
wilddamon: It still shouldn't be inside a second array though? Just [ parsedThing, remaining ]. If something else needs this to be in an array it should do it there.

test/js/css-url-image-value.js
```diff
@@ -18,32 +18,56 @@ suite('CSSURLImageValue', function() {
...
+      function() {
+        assert.strictEqual(urlImageValue.state, "loaded");
+        assert.strictEqual(urlImageValue.intrinsicWidth, 1);
+        assert.strictEqual(urlImageValue.intrinsicHeight, 1);
+        assert.strictEqual(urlImageValue.intrinsicRatio, 1);
+        done();
+      }
+    ]);
+
+    // urlImageValue._image.onload = function() {
```
wilddamon: Don't forget to remove this.

wilddamon:
LGTM except for minor nit. Let me know when it's done and I'll merge.

test/js/css-url-image-value.js
anthony-huang: Done :)

src/parsing/css-url-image-value-parsing.js
anthony-huang: Done :)

wilddamon:
LGTM

PR #145, opened 2016-09-02
CSSVariableReferenceValue's fallback can be undefined

anthony-huang:
In this example https://drafts.css-houdini.org/css-typed-om/#cssunparsedvalue the fallback can be undefined.


wilddamon:
LGTM

Issue #88, opened 2016-06-28
Many tests fail using npm test due to "Can't find variable: CSS"


wilddamon:
In CSSKeywordValue, we call CSS.escape, which doesn't seem to be available in node.

samthor:
Closing in favor of #152, which is the new reason it's broken.

PR #150, opened 2016-09-12
adds empty CSS.escape polyfill, needed ES6 polyfills

samthor:
This allows `npm test` to run, and most tests now pass.

Depending on choice of ES6 transpiler, and the options we use, the `Number` and `String.prototype` polyfills might eventually come for free.


samthor:
Note that this would resolve #88.

wilddamon:
LGTM

PR #155, opened 2016-09-14
Make style used for hiding CSSImageValue's constructor consistent with other code

wilddamon:
Also update tests.

### DIFF
--- a/src/css-image-value.js
+++ b/src/css-image-value.js
@@ -45,17 +45,13 @@
       }
 
       this._image = image;
-      //if (image.complete) {
-      //  onLoad.call(this);
-      //} else {
-        this.state = "unloaded";
-        this.intrinsicWidth = null;
-        this.intrinsicHeight = null;
-        this.intrinsicRatio = null;
-        this._image.onload = onLoad.bind(this);
-        this._image.onerror = onError.bind(this);
-        this._image.onprogess = onProgress.bind(this);
-      //}
+      this.state = "unloaded";
+      this.intrinsicWidth = null;
+      this.intrinsicHeight = null;
+      this.intrinsicRatio = null;
+      this._image.onload = onLoad.bind(this);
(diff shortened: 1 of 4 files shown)


src/css-image-value.js
```diff
@@ -14,48 +14,47 @@
...
-  (function() {
-    function CSSImageValue() {
-      throw new TypeError("Can\'t instantiate CSSImageValue");
-    }
-    internal.inherit(CSSImageValue, CSSResourceValue);
+  function onLoad() {
+    this.state = "loaded";
+    this.intrinsicWidth = this._image.naturalWidth;
+    this.intrinsicHeight = this._image.naturalHeight;
+    if (this.intrinsicHeight != 0)
```
samthor: { } around single-line statements for consistency

src/css-image-value.js
```diff
@@ -14,48 +14,47 @@
...
+      if (!(image instanceof Image)) {
+        throw new TypeError("image must be an Image object");
+      }
+
+      this._image = image;
+      this.state = "unloaded";
+      this.intrinsicWidth = null;
+      this.intrinsicHeight = null;
+      this.intrinsicRatio = null;
+      this._image.onload = onLoad.bind(this);
```
samthor, line 52: This is unrelated to this patch, but I don't think this does what you want. If the `Image` is already loaded, then this will never be called.

samthor:
LGTM, but note the `onLoad` problem

src/css-image-value.js
wilddamon: Done

src/css-image-value.js
wilddamon: Ok, I think I saw you put in a patch to address this too :)

wilddamon:
Thanks!

wilddamon:
This review thing is weird in an already open PR, but cool!

PR #157, opened 2016-09-14
Hide CSSUnparsedValue's constructor

wilddamon:
It shouldn't be instantiable by users.

### DIFF
--- a/src/css-unparsed-value.js
+++ b/src/css-unparsed-value.js
@@ -53,8 +53,7 @@
     function CSSUnparsedValue(values) {
       if (values == undefined) {
         values = [];
-      }
-      if (!Array.isArray(values)) {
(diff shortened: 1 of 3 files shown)


src/css-unparsed-value.js
```diff
@@ -60,4 +49,24 @@
 
   scope.CSSUnparsedValue = CSSUnparsedValue;
 
+  (function() {
+    function CSSUnparsedValue(values) {
+      if (values == undefined) {
+        values = [];
+      }
```
samthor: nit: could be `else if` since the next branch will never run

samthor:
LGTM

src/css-unparsed-value.js
wilddamon: Done.

PR #158, opened 2016-09-15
refactor css-image-value to copy src, not use passed Image

samthor:
I'm not quite sure this is the semantics you want. But the problem is that it's hard to tell whether a passed `Image`, which already has its `.src` property set, has an error state.

Basically you can check its [complete](https://developer.mozilla.org/en/docs/Web/API/HTMLImageElement) property, but this just indicates loaded. There's no recorded error state, so the only way to find out whether it fails is by creating it anew or setting its `.src` property to reload.

Presumably you want to also eventually expose the `Image` to users of `CSSImageValue`.

### DIFF
--- a/src/css-image-value.js
+++ b/src/css-image-value.js
@@ -40,18 +40,20 @@
 
   (function() {
     function CSSImageValue(image) {
-      if (!(image instanceof Image)) {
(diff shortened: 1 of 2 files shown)


wilddamon (approved):
This seems fine, especially since it's an internal object.

